PR #631, opened 2026-04-15
fix: resolve v2.6.123 crash-loop via @lucky/shared barrel import

LucasSantana-Dev:
PRODUCTION HOTFIX — bot was crashing on startup with ERR_PACKAGE_PATH_NOT_EXPORTED. Patched #629 added the utils/* subpath export but Node couldn't resolve extension-less subpath import of errorSanitizer. Fix: import via the @lucky/shared/utils barrel which already uses '.' mapping.


LucasSantana-Dev:
Superseded by #633 — the real root cause was Dockerfile not copying packages/bot/node_modules (where workspace-pinned lru-cache@10 lives). Import paths were fine, the module was just missing from the runtime image.

PR #729, opened 2026-04-20
feat(backend): top.gg vote webhook endpoint

LucasSantana-Dev:
## Summary
Adds \`POST /webhooks/topgg-votes\` to persist vote events for the upcoming top.gg listing (Phase 0 of the next-phases ultraplan).

Locked behind \`TOPGG_AUTH_TOKEN\` — returns 503 when unset, so the endpoint is inert until the listing actually goes live.

### Behaviour
- Verifies the \`authorization\` header against \`TOPGG_AUTH_TOKEN\` (top.gg's webhook auth scheme).
- For \`type: "test"\` payloads → returns \`{ ok: true, test: true }\` (lets top.gg's "Send test webhook" button pass).
- For \`type: "upvote"\`:
  - \`votes:<userId>\` — 12h TTL (matches top.gg's 12h vote window)
  - \`votes:streak:<userId>\` — incremented, 36h TTL (12h grace around the 24h cycle so a daily voter doesn't lose their streak to timezone drift)
  - Uses a single Redis pipeline so either all three ops succeed or we return 500.

### Wiring
- Registered in \`routes/index.ts\` alongside the other non-API endpoints (health, stats, internal notify).
- **No \`/api/\` prefix** — matches top.gg's expected webhook URL shape and bypasses the per-origin \`apiLimiter\`. Still rate-limited via \`writeLimiter\`.

### Env
- \`TOPGG_AUTH_TOKEN\` added to \`.env.example\` (commented out).

## Test plan
- [x] \`npm run type:check --workspace=packages/backend\` → clean
- [x] 6 unit tests cover: missing token (503), missing/wrong auth (401), test-type payload, missing user id (400), successful upvote (verifies all 3 pipeline ops + correct TTLs).
- [ ] After merge + deploy: run top.gg's "Send test webhook" button; expect 200 with \`{ ok: true, test: true }\`
- [ ] Real upvote writes two Redis keys with correct TTLs (manually verifiable via \`redis-cli\`)

## Follow-up
Next PR will add \`/voterewards\` slash command in \`packages/bot\` that reads \`votes:streak:<user>\` and grants tier benefits. See \`docs/TOP_GG_SUBMISSION.md\` §6 for the broader wiring plan.

## Summary by CodeRabbit

* **New Features**
  * Added top.gg vote webhook and public/internal endpoints to record and report votes
  * Added `/voterewards` Discord command showing vote status, streak, current/next tier, and vote link
  * Exposed vote tiers and vote URL used by UI and commands

* **Tests**
  * Added unit tests covering webhook endpoints and the voterewards command

* **Chores**
  * Added optional TOPGG_AUTH_TOKEN entry and docs to environment example

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -198,3 +198,8 @@ SEARCH_RETRY_DELAY=5000
 # DEPENDENCY_CHECK_INTERVAL=86400000
 # DEPENDENCY_NOTIFY_ONLY_SECURITY=false
 LUCKY_NOTIFY_API_KEY=
+
+# top.gg vote webhook (optional — for the /webhooks/topgg-votes endpoint)
+# Get this from https://top.gg/bot/<BOT_ID>/webhooks after listing the bot.
+# The value you paste in top.gg's "Authorization" field must match this.
+# TOPGG_AUTH_TOKEN=your_topgg_webhook_auth_token
--- a/packages/backend/src/routes/index.ts
+++ b/packages/backend/src/routes/index.ts
@@ -18,6 +18,7 @@ import { setupStarboardRoutes } from './starboard'
 import { setupMusicRoutes } from './music'
 import { setupArtistsRoutes } from './artists'
 import { setupInternalNotifyRoutes } from './internalNotify'
+import { setupWebhookRoutes } from './webhooks'
 import { apiLimiter } from '../middleware/rateLimit'
 import { requireAuth } from '../middleware/auth'
 import { requireGuildModuleAccess } from '../middleware/guildAccess'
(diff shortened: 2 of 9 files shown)


LucasSantana-Dev:
Addressed all three substantive CodeRabbit findings in 6604adc6:

- ✅ **Idempotent upvote handler** — SET NX gates the streak increment; retries return `{ ok: true, duplicate: true }` without touching the counter
- ✅ **Timing-safe token compare** — both `TOPGG_AUTH_TOKEN` and `LUCKY_NOTIFY_API_KEY` now use `crypto.timingSafeEqual` via a length-checked helper
- ✅ **Redis error listener** — attached to the cached client so connection failures don't raise uncaughtException
- ✅ **Explicit `type === 'upvote'` guard** — unknown/future event types now 400 instead of silently persisting

Tests: added 2 new cases (duplicate vote idempotency, unsupported type rejection) + updated the existing "records upvote" assertion to check SET NX. 15/15 passing.

The `interactionReply` defer and rate-limit middleware queries were investigative, not actionable — left as-is.

PR #745, opened 2026-04-20
docs(spec): Phase 2 premium-tier foundation

LucasSantana-Dev:
## Summary
Turns the vague \"L-effort greenfield\" in the next-phases ultraplan into a concrete 4-PR plan with schema, endpoints, sequencing, and rollback strategy. **Zero code, zero runtime impact** — pure planning artifact.

### Key decisions (see \`spec.md\` for full rationale)
- **Single \$2.99/mo tier for MVP** — undercuts Rythm (\$4.99), premium-feeling vs. MEE6 entry (\$1.99). Tier ladder deferred to post-gate ($50 MRR target).
- **New Prisma models**: \`GuildSubscription\` mirrors Stripe lifecycle; \`StripeWebhookEvent\` idempotency by \`event.id\`.
- **Status is authoritative from Stripe**; \`currentPeriodEnd\` informational only (avoids clock-skew bugs).
- **Three initial feature gates** chosen for minimum blast radius: premium autoplay depth (cap 20 → uncapped), custom embed color, \`/leaderboard --limit > 10\`. All fail gracefully with dashboard link.
- **Every PR behind \`STRIPE_ENABLED\` env flag** — rollback is a config flip.
- **Bot reads subscription via backend HTTP** (same internal-key pattern as vote-status) — no Postgres drift.

### 4-PR sequencing (~16h total)
| PR | Effort | Contents |
|---|---|---|
| 1 | S (~2h) | Schema + migration + \`PremiumService.isPremium()\` stub returning false |
| 2 | M (~6h) | Stripe SDK + billing routes + webhook handler with idempotency |
| 3 | S (~2h) | First gate: premium autoplay depth (20 → uncapped) |
| 4 | M (~6h) | Dashboard \`/settings/billing\` + 2 more gates (embed color, leaderboard limit) |

## Test plan
- [x] Spec validates (frontmatter parses, `status: proposed`)
- [x] Pricing anchor cross-referenced against competitive-analysis doc
- [x] All 4 PRs have acceptance criteria in tasks.md
- [ ] After merge: link this spec from next-phases-ultraplan as the Phase 2 implementation plan

## Next actions
- Review/tweak in discussion if needed
- When ready to execute: create the 4 PRs sequentially following tasks.md. Each has its own acceptance criteria.

PR #749, opened 2026-04-20
feat(backend): Stripe billing route scaffolding (503 when disabled)

LucasSantana-Dev:
## Summary
Introduces the \`/api/billing/*\` and \`/webhooks/stripe\` route surface gated behind \`STRIPE_ENABLED\`. **Zero runtime change** — \`STRIPE_ENABLED\` defaults to unset so production hits the 503 path everywhere.

Sets the shape for the real billing PR without touching the \`stripe\` npm dep, Prisma, or the frontend.

### Added
- **\`utils/billingConfig.ts\`** — \`isStripeEnabled()\`, \`getStripeConfig()\`, \`requireStripeEnabled()\`. Tiny env-reading helpers, no caching. Pattern mirrors \`utils/developerAccess.ts\`.
- **\`routes/billing.ts\`** — 4 endpoints (status GET, checkout POST, portal POST, subscription DELETE). Each short-circuits via \`requireStripeEnabled()\` then throws \`AppError(501, 'Not implemented')\` so enabling the flag later produces predictable "to be implemented" responses.
- **\`routes/stripeWebhook.ts\`** — \`POST /webhooks/stripe\`. Top-of-file comment flags the \`express.json()\` body-parsing issue that will matter once real signature verification lands (not solved here because the 503 returns before body access).
- **\`routes/index.ts\`** — imports + \`routeSetups\` entries for both new modules.

### Wiring notes
- \`/api/billing/*\` inherits \`apiLimiter\` via the existing \`/api/\` prefix middleware.
- Billing mutations use \`writeLimiter\`.
- \`/webhooks/stripe\` is deliberately non-\`/api/\` so it skips \`apiLimiter\` (Stripe retries on its own cadence; client-IP rate limiting would be counterproductive).
- No \`requireAuth\`, no \`guildGuardConfigs\` — those are future-PR concerns and the 503 returns before they'd apply.

### Follow-ups (explicitly NOT in this PR)
- \`stripe\` npm dep
- Prisma subscription reads via \`PremiumService\` (#746)
- Real checkout/portal session creation
- Webhook signature verification + body-parsing fix

## Test plan
- [x] \`npm run type:check --workspace=packages/backend\` → clean
- [x] 8 unit tests pass: 4 \"disabled → 503\" (one per billing endpoint) + \`STRIPE_ENABLED=false\` case + \`STRIPE_ENABLED=true → 501\` spot-check + 2 matching cases on the webhook
- [ ] Optional smoke: \`curl -i localhost:5000/api/billing/status\` → 503, then with \`STRIPE_ENABLED=true\` → 501

## Sequencing
Builds on #747 (env placeholders), #746 (\`PremiumService\` + schema), #745 (spec).

### DIFF
--- a/packages/backend/src/routes/billing.ts
+++ b/packages/backend/src/routes/billing.ts
@@ -1,0 +1,46 @@
+import type { Express, Request, Response } from 'express'
+import { writeLimiter } from '../middleware/rateLimit'
+import { asyncHandler } from '../middleware/asyncHandler'
+import { AppError } from '../errors/AppError'
+import { requireStripeEnabled } from '../utils/billingConfig'
+
+// Route surface only — real handlers land in the next PR. Every handler
+// short-circuits via requireStripeEnabled() so production (with
+// STRIPE_ENABLED unset) sees 503 here.
+
+export function setupBillingRoutes(app: Express): void {
+    app.get(
+        '/api/billing/status',
+        asyncHandler(async (_req: Request, _res: Response) => {
+            requireStripeEnabled()
+            throw new AppError(501, 'Not implemented')
+        }),
+    )
+
+    app.post(
+        '/api/billing/checkout',
+        writeLimiter,
+        asyncHandler(async (_req: Request, _res: Response) => {
+            requireStripeEnabled()
+            throw new AppError(501, 'Not implemented')
+        }),
+    )
+
+    app.post(
+        '/api/billing/portal',
+        writeLimiter,
+        asyncHandler(async (_req: Request, _res: Response) => {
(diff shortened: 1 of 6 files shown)


LucasSantana-Dev:
Paused: Stripe/premium-billing work deferred. The existing `premiumService.isPremium(guildId)` in @lucky/shared already acts as the feature toggle (returns `false` for every guild without a subscription row). Re-open or re-create when premium work resumes. Related: #746 (schema already on main, harmless no-op), #750 (premium autoplay — gate works as the feature flag).

PR #1077, opened 2026-05-28
fix(deps): sync package-lock.json – unblock Vercel @types/react build failure

LucasSantana-Dev:
## Problem

Vercel builds have been failing since commit `34c309f8` (prom-client root dep hoisting) with:

```
npm error `npm ci` can only install packages when your package.json and
package-lock.json or npm-shrinkwrap.json are in sync.
npm error Missing: @types/react@19.2.15 from lock file
```

## Root cause

The lockfile had `@types/react@19.2.15` stored at the workspace-local path (`packages/frontend/node_modules/@types/react`) rather than the root (`node_modules/@types/react`). When `34c309f8` ran `npm install` to add prom-client, npm's resolution algorithm changed the expected location (hoisting), but the lockfile structure didn't reflect this. `npm ci` enforces strict sync and fails when the lockfile diverges from what `npm install` would produce.

## Fix

Ran `npm install` from repo root to regenerate `package-lock.json` with the correct hoisted structure. `@types/react@19.2.15` is now at `node_modules/@types/react` (root-hoisted) and the workspace-local duplicate entry has been removed.

## Verification

- `node_modules/@types/react: 19.2.15` — present in lockfile ✓
- `packages/frontend/node_modules/@types/react` — removed (no longer needed) ✓
- Pre-commit type-check passed across all packages ✓

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1859,7 +1859,7 @@
             "version": "0.5.0",
             "resolved": "https://registry.npmjs.org/@clack/core/-/core-0.5.0.tgz",
             "integrity": "sha512-p3y0FIOwaYRUPRcMO7+dlmLh8PSRcrjuTndsiA0WAFbWES0mLZlrjVoBRZ9DzkPFJZG6KGkJmoEAY0ZcVWTkow==",
-            "dev": true,
+            "devOptional": true,
             "license": "MIT",
             "dependencies": {
                 "picocolors": "^1.0.0",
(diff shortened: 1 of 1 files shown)


LucasSantana-Dev:
Redundant: the @types/react lockfile fix is already on main via #1080 (v2.15.2 lockfile regen — `@types/react` at root, no workspace-local dup). Rebase would yield a no-op. Closing.

PR #1388, opened 2026-06-13
fix(bot): thread real Client into endGiveaway (#1383)

### LINKED ISSUE #1383: giveaway end/reroll silently no-ops: require('../../../client') resolves to a non-existent module
## Summary
`endGiveaway` in `packages/bot/src/functions/general/commands/giveaway.ts` lazily loads the bot client via `require('../../../client').default` when no `client` arg is passed. That path resolves to `packages/bot/src/client`, **which does not exist** — the client is constructed in `packages/bot/src/handlers/clientHandler/service.ts`, and there is no `src/client.ts` / `src/client/index.ts` / default-client export anywhere. So the `require` throws, the surrounding `try/catch` swallows it, and the function returns early.

## Why it's always hit
All three call sites omit the `client` argument, so every call falls into the `if (!client)` fallback branch:
- `giveaway.ts:270` — `endGiveaway(message.id)` (setTimeout completion)
- `giveaway.ts:310` — `await endGiveaway(messageId)` (manual end)
- `giveaway.ts:346` — `await endGiveaway(messageId, true)` (reroll)

## Impact
Giveaway completion and reroll never run — no winner is picked/announced. Silent failure (the `catch` returns without logging).

## Evidence
- `giveaway.ts:64-69`:
  ```ts
  if (!client) {
      try {
          client = require('../../../client').default
      } catch {
          return
      }
  }
  ```
- `find packages/bot/src -name 'client*'` → only `events/clientReady.ts`, `handlers/clientHandler*` (no `src/client`).
- Reproduces on clean `main` (pre-existing; surfaced while clearing no-unsafe lint for #1378).

## Suggested fix
Pass the client through from the call sites (the interaction/timeout context already has `interaction.client` / the started client), or export a real client singleton and import it. Then the `any`-typed `require` (currently a `no-unsafe-*` site under #1378) goes away too. At minimum, log on the catch instead of silently returning.

LucasSantana-Dev:
## Root Cause
`endGiveaway` in `packages/bot/src/functions/general/commands/giveaway.ts` had a broken fallback path that attempted to require a non-existent client module at `../../../client` when no client argument was provided. The require would throw, the surrounding try/catch would suppress the error, and the function would return early without processing giveaway completion or rerolls.

All three call sites omitted the client argument, so winner selection and announcement never ran.

## Fix
Thread the real `Client` instance from each call site to `endGiveaway`:
- **Line 273 (setTimeout completion):** Pass `interaction.client` from the surrounding command handler
- **Line 313 (manual end):** Pass `interaction.client` from the 'end' subcommand handler
- **Line 349 (reroll):** Pass `interaction.client` from the 'reroll' subcommand handler

Removed the broken fallback require entirely since all call sites now supply a client.

## Verification
- ✅ `npm run build --workspace=packages/bot` — compiles successfully
- ✅ `npm run lint --workspace=packages/bot` — no unsafe/require warnings on giveaway.ts; only pre-existing complexity warning remains
- ✅ No breaking changes; purely fixes silent failure mode

Closes #1383

---
## Summary by cubic
Fixes giveaways not completing or rerolling by passing the real Discord `Client` into `endGiveaway`. Removes a broken fallback that suppressed errors and returned early.

- **Bug Fixes**
  - Pass `interaction.client` to `endGiveaway` for timeout completion, manual end, and reroll.
  - Remove the `require('../../../client')` fallback since all call sites now supply a client.

<sup>Written for commit 25bcf630469232a5f0afcf2b376b165d7a93e956. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/1388?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/bot/src/functions/general/commands/giveaway.ts
+++ b/packages/bot/src/functions/general/commands/giveaway.ts
@@ -61,17 +61,6 @@ async function endGiveaway(
 
     if (!giveaway) return
 
-    if (!client) {
-        try {
-            // NOTE: this fallback path is currently broken — `../../../client`
-            // resolves to a non-existent module, so the require throws and the
-            // catch returns. Tracked in #1383. Cast keeps the type honest.
-            client = (require('../../../client') as { default: Client }).default
-        } catch {
-            return
-        }
-    }
-
     if (!client) return
     const guild = client.guilds.cache.get(giveaway.guildId)
     if (!guild) return
@@ -270,7 +259,7 @@ export default new Command({
                     endTime,
                     entries: new Set(),
                     timeoutId: setTimeout(() => {
-                        endGiveaway(message.id)
+                        endGiveaway(message.id, false, interaction.client)
                     }, durationMs),
                 }
 
@@ -310,7 +299,7 @@ export default new Command({
                     return
                 }
 
-                await endGiveaway(messageId)
+                await endGiveaway(messageId, false, interaction.client)
 
                 await interactionReply({
                     interaction,
@@ -346,7 +335,7 @@ export default new Command({
                     return
                 }
 
-                await endGiveaway(messageId, true)
+                await endGiveaway(messageId, true, interaction.client)
 
                 await interactionReply({
                     interaction,


LucasSantana-Dev:
@cubic-dev-ai please review for correctness

PR #1396, opened 2026-06-13
feat(backend): move session store from Redis to Postgres (#1111)

### LINKED ISSUE #1111: Migrate Redis KV/cache → Postgres + in-memory (keep Redis for music pub/sub)
## Problem Statement

As the operator of a single-instance hobby music bot, I run a Redis container (256MB) that ~22 modules depend on for caching and storage, plus a Postgres database (Prisma) for relational data. For this scale, Redis is doing work that Postgres (already running) or in-memory maps could do — adding an infra dependency, a 256MB container, and a startup hard-dependency (the bot throws if Redis is down) for little benefit. I want to reduce Redis to only what genuinely needs it.

## Solution

Per ADR `docs/decisions/2026-05-31-redis-scope-reduction.md` (Option A — "decouple", critic-validated): migrate Redis's **KV/cache** role off Redis, and **retain Redis only for the `MusicControlService` bot↔backend pub/sub** (`music:command|result|state`, which powers the web music-control surface). Persistent stores move to **Postgres/Prisma**; ephemeral TTL caches move to **in-memory** maps.

This is delivered **incrementally, one store per PR**, behind a **Phase-0 prototype** (`TrackHistory`) that validates the Postgres TTL + indexing pattern before the rest proceed. Dropping Redis *entirely* (re-architecting the music IPC) is explicitly out of scope here.

## User Stories

1. As the operator, I want guild track history to survive bot restarts/deploys, so that `/history` and autoplay stats aren't wiped on every release.
2. As the operator, I want named saved queues to persist in Postgres, so that a deploy doesn't lose a user's saved queue.
3. As the operator, I want autoplay/repeat counters to persist, so that autoplay behavior is stable across restarts.
4. As the operator, I want music session-restore snapshots to persist, so that an interrupted session can resume after a restart.
5. As the operator, I want ephemeral caches (provider-health, AutoMod spam windows, short-lived settings caches) to live in-memory, so that they add no infra dependency and simply rebuild after a restart.
6. As the operator, I want the bot to NOT hard-fail at startup when Redis is unavailable for KV purposes, so that only the pub/sub path depends on Redis.
7. As the operator, I want each migrated store to keep its existing service interface, so that callers (routes, handlers, frontend API) don't change.
8. As the operator, I want each Postgres-backed store to have an explicit expiry strategy, so that old rows don't accumulate unbounded.
9. As the operator, I want the `TrackHistory` store migrated first as a prototype, so that the migration pattern is proven before committing the rest.
10. As a developer, I want the Postgres-backed stores covered by behavior tests (write→read, ordering, cap, expiry), so that the migration is verifiable without a live Redis.
11. As the operator, I want the watchdog orphan-scan to use an indexed query instead of a Redis `keys('music:session:*')` scan, so that it doesn't degrade as guild count grows.
12. As the operator, I want Redis removed from `@lucky/shared`'s KV consumers so that only the pub/sub service imports `ioredis`.

## Implementation Decisions

- **Stores → Postgres/Prisma (deep modules, interface preserved):**
  - `TrackHistory` — **Phase-0 prototype.** The `TrackHistory` Prisma model **already exists** (`@@index([guildId, playedAt])`, FK to `Guild`, cascade delete) and maps cleanly to the existing `TrackHistoryEntry` interface, so **no schema migration is required** — only rewiring `TrackHistoryService` storage methods from Redis to `prisma.trackHistory`.
  - `NamedQueueStore` (named saved queues), `GuildCounterStore` (autoplay/repeat counters), `SessionSnapshotStore` (music session-restore). Each gets a Prisma model + indexed `guildId`.
- **Stores → in-memory (ephemeral):** `ProviderHealthCache`, AutoMod spam-window tracking, short-lived settings caches. Backed by a `Map` + TTL; lost-on-restart is acceptable.
- **TTL strategy:** lazy-expiry-on-read (filter `WHERE expiry/playedAt >= cutoff`) as the primary mechanism, with an optional periodic sweep (`deleteMany WHERE ... < cutoff`) for batch-read stores. Per-store choice documented.
- **Cap strategy (e.g. `TrackHistory` 100-entry cap):** enforced on write by trimming rows beyond the newest N for that guild.
- **Watchdog orphan-scan:** replace `redisClient.keys('music:session:*')` with an indexed Postgres query on the snapshot table's `guildId`.
- **Startup:** the bot's Redis hard-dependency at startup becomes pub/sub-only; KV unavailability degrades gracefully.
- **Retained on Redis (out of scope):** `MusicControlService` pub/sub (`music:command|result|state`).
- **Sequencing:** one store per PR. Phase-0 (`TrackHistory`) gates the rest — if it exposes >3 unanticipated friction points or needs >2 shims, escalate to `/research-and-decide` before continuing.

## Testing Decisions

- Good tests assert **external behavior**, not implementation: given writes, the store returns the right reads (ordering, limit/offset, cap, expiry-filter, clear), independent of the storage backend.
- **Modules tested:** the Postgres-backed stores (`TrackHistory` first, then `NamedQueueStore`, `GuildCounterStore`, `SessionSnapshotStore`). The trivial in-memory caches are not separately unit-tested (their behavior is the `Map`'s).
- **Prior art:** existing `packages/shared` service specs using ts-jest with a factory-mocked `prismaClient` and `@jest/globals` imports (`jest.fn<Fn>()` single-type-arg form), run under the shared package's strict tsconfig via `Quality Gates`.
- Each store's spec must pass without a live Redis or Postgres (mock the Prisma client).

## Out of Scope

- **Removing Redis entirely / re-architecting the music pub/sub** (Postgres `LISTEN/NOTIFY` or WebSocket) — that is ADR Option B, deferred as its own gated decision.
- Backend HTTP session store (`connect-redis`) — already has a file/memory fallback; not part of this migration.
- Multi-instance support — the in-memory caches assume single-instance (a revisit trigger in the ADR).

## Further Notes

- Decision record: `docs/decisions/2026-05-31-redis-scope-reduction.md`.
- The Opus `critic` flipped the original "drop Redis entirely" framing; this PRD reflects the revised, decoupled scope.
- Phase-0 (`TrackHistory`) is intentionally the smallest representative store (list + TTL + cap) to prove the pattern cheaply.

LucasSantana-Dev:
## Why

Closes the final item of the Redis scope-reduction arc (ADR `2026-05-31-redis-scope-reduction`, #1111). `express-session` was the **last KV consumer still on Redis** (via `connect-redis`). After this, Redis is retained **only** for the `MusicControlService` pub/sub, exactly as the ADR specifies.

## Approach

Chose a **Postgres-backed `PrismaSessionStore` that reuses the existing Prisma client** over `connect-pg-simple`. Rationale:
- `connect-pg-simple` would require adding `pg` and a **second connection pool** to Supabase — cutting against the whole point of the arc (reducing infra/connection dependencies) and Supabase's pooled-connection limits.
- The repo's consistent precedent (every Redis-reduction ADR) is *read Postgres direct via the existing Prisma client*. A ~120-line store adapter reuses that client with **zero new runtime deps and no extra pool**.

## Changes

- **Schema + migration**: new `sessions` table (`model Session`) with an index on `expiresAt` backing the prune sweep. Migration `20260613060000_add_sessions_table` verified by `prisma-migration-verifier` (schema/migration/client parity, additive-only, indexed, no FK).
- **`PrismaSessionStore`**: `get`/`set`/`destroy`/`touch` with lazy expiry on read + an **unref'd** periodic prune. P2025 (record-not-found) treated as no-op success; corrupt payloads treated as no-session rather than wedging auth.
- **Resilience preserved**: still wrapped by the existing `ResilientSessionStore`, so a Postgres blip fails over to the local file store — same contract the Redis primary had.
- **Cleanup**: removed the `connect-redis` dependency and the dead Redis session adapter (`createConnectRedisClientAdapter` + its now-obsolete test).
- **Tests**: 15 new unit cases for `PrismaSessionStore`; updated the session-middleware test for the Postgres path.

## Verification

- Full backend suite: **67 suites / 1010 tests green** (incl. the session-touching integration tests).
- Type-check green across shared/backend/bot/frontend (pre-commit).
- Migration verified read-only by `prisma-migration-verifier` → PASS.

## Deploy note

The `sessions` table is created by the migration (`prisma migrate deploy` in the deploy chain). No env-var changes. Existing Redis-stored sessions are not migrated — users get a fresh login after deploy (acceptable for a dashboard session store).

Closes #1111.

---
## Summary by cubic
Moved `express-session` from Redis to a Postgres-backed `PrismaSessionStore` to remove the last Redis KV usage and simplify infra. Aligns with ADR 2026-05-31-redis-scope-reduction; Redis now remains only for `MusicControlService` pub/sub (closes #1111).

- **Refactors**
  - Added `PrismaSessionStore` (get/set/destroy/`touch`) with lazy expiry and an unref’d hourly prune.
  - Kept resilience via `ResilientSessionStore` to fail over to the local file store on DB issues.
  - `touch` now propagates non-`P2025` errors so failover triggers; added a regression test.
  - Removed `connect-redis` and the Redis adapter/tests; cleaned up a dead callback type export and a duplicate `@lucky/shared/utils` import; updated session middleware and unit tests.

- **Migration**
  - Added `Session` model and `sessions` table with an index on `expiresAt`.
  - No env changes; run `prisma migrate deploy`. Redis sessions aren’t migrated; users will re-login.

<sup>Written for commit d82076b0292ef3df2a65e6bc5508f306c5bfc746. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/1396?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LucasSantana-Dev:
@cubic-dev-ai please review

LucasSantana-Dev:
@cubic-dev-ai addressed both findings in d82076b0:

- **P1 (touch swallows all errors):** valid — fixed. `touch` now mirrors `destroy`: only `P2025` (missing row) is a benign no-op; all other DB errors propagate so `ResilientSessionStore` can fail over on a Postgres outage instead of silently reporting success.
- **P2 (missing test):** added a regression test asserting a non-`P2025` error reaches the `touch` callback.

Store suite now 16/16 green.

Issue #1397, opened 2026-06-14
chore(deploy): rapid successive merges cause false-failure in 'Validate deployed version'


LucasSantana-Dev:
## Observed

Two merges landed ~16 min apart on 2026-06-13:
- `78a30f31` (fix/twitch #1395) merged 23:57 → "Deploy to Homelab" run [27482856341](https://github.com/LucasSantana-Dev/Lucky/actions/runs/27482856341) **FAILED**
- `ff5e0b68` (session-store #1396) merged 00:13 → run [27483193936](https://github.com/LucasSantana-Dev/Lucky/actions/runs/27483193936) **succeeded**

The failed run's `Validate deployed version` step polled 20×/15s (~5 min) and the homelab **never** advanced to `78a30f31` — it stayed on the prior commit `19723a70` the whole time, then errored: `Deployed version SHA did not match expected after all attempts`.

## Analysis

Net state is **correct**: prod ended on `ff5e0b68` (which contains `78a30f31`'s changes + the sessions migration), and that deploy validated green. So no change was lost. But the earlier deploy run goes **RED on a benign supersession** — the homelab was already moving to the newer commit, so the older run's expected SHA could never match.

**Why it matters:** this false-failure is indistinguishable from a *real* failed deploy in the Actions UI/notifications — alarm fatigue that can mask a genuine deploy break. It recurs whenever two merges land within the deploy/validate window.

## Options (needs decision)

1. **Concurrency group with cancel-in-progress** on the deploy workflow (`concurrency: { group: deploy-prod, cancel-in-progress: true }`) so a newer merge cancels the superseded deploy instead of letting it fail validation. ⚠️ Must ensure cancellation can't interrupt an in-flight `prisma migrate deploy` mid-statement.
2. **Tolerate supersession in validation**: if the actual deployed SHA is a *descendant* of the expected SHA on `main` (i.e. a newer merge superseded this one), exit 0 as "superseded" rather than failing.

Option 2 is lower-risk (no deploy interruption). 

## Evidence
- Fail log: run 27482856341, step `Validate deployed version` (attempts 1–20, all `SHA mismatch (expected=78a30f31..., actual=19723a70...)`).
- Success: run 27483193936 (`ff5e0b68`, conclusion success).

Surfaced during post-merge deploy verification.

LucasSantana-Dev:
[...]

PR #1398, opened 2026-06-14
ci(deploy): gate production deploy to release publication, not every main push

LucasSantana-Dev:
Production currently ships on **every push to `main`** — `docker-publish` builds images on each main commit and **Deploy to Homelab** chains off it via `workflow_run`, so every merge auto-deploys to prod (and they race — see #1397). This makes shipping accidental rather than deliberate.

## Change
Gate production deploys to **GitHub Release publication**:

- **Trigger:** swap `workflow_run` (Build & Push Docker Images / main) → `release: types: [published]`. `workflow_dispatch` (manual deploy + rollback) is unchanged.
- **SHA resolution:** add one `resolve_sha` step as the single source of truth for the shipped commit. On a release it dereferences the **annotated** tag to its commit SHA (`github.sha` is unreliable on release events). Rollback and `workflow_dispatch` paths are preserved.
- **Consistency:** every downstream step (image wait, webhook, Sentry create/finalize, version validation, homelab wait) now reads that one resolved SHA, so the Sentry release, the deployed `:<sha>` image, and the validated commit status can never disagree (they were independently re-derived in 6 places before).

## What does NOT change
- `docker-publish.yml` still builds `:<sha>` + `:latest` on every main push, so the image for a release commit already exists when the release is cut. **Building artifacts is decoupled from shipping them.**
- The `Production` environment protection (wait timer) still applies — now on release deploys.
- Rollback via `workflow_dispatch` `rollback_sha` is untouched.

## Shipping flow after this
`merge to main` → images built (no deploy) → cut release `vX` (push annotated tag → `Release` workflow publishes it) → **Deploy to Homelab fires for the tagged commit**.

## Known limitation
If a release tag points at a commit that never built a `:<sha>` image (e.g. a docs-only HEAD that `docker-publish` path-filtered out), the deploy falls to the existing `:latest` path. In practice release tags sit on code/`package.json` commits that always build. Flagging for review rather than over-engineering.

## Verification
- `actionlint`: 0 new findings vs `main` baseline (7 pre-existing `SC2016` info notes on intentional single-quoted `jq`/`node -e`, identical on both).
- YAML parses; no `workflow_run` / stray `github.sha` references remain outside the intended `workflow_dispatch` path.
- Pre-commit type-check (shared/bot/backend/frontend) passed.

@cubic-dev-ai please review.

---
## Summary by cubic
Gate production deploys to GitHub Release publication instead of every push to `main`. Adds SHA validation and env-based propagation to prevent injection, keeping Sentry, image tags, and commit status in sync, and improves Docker build lookup for older tags.

- **Refactors**
  - Trigger changed: `workflow_run` → `release: [published]`; `workflow_dispatch` (manual/rollback) stays.
  - Added `resolve_sha` to pick the shipped commit once; on release, dereferences the annotated tag to its commit SHA.
  - Validates the resolved SHA (7–40 lowercase hex) and passes it via env; all steps read it to prevent template-injection via `rollback_sha`.
  - Docker wait now finds the build by exact `head_sha` via the Actions API (not a recent window), so releases for older tags don’t fail lookup.
  - Downstream steps (image wait, webhook, Sentry create/finalize, version validation, homelab wait) use the one SHA.
  - `docker-publish.yml` still builds `:<sha>` and `:latest` on `main`; building artifacts is decoupled from shipping them.

<sup>Written for commit 0b4902a2712a99fa4edfd5aef51f9f8a9b97cdc4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/1398?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Chores**
  * Updated deployment workflow to trigger on release publication instead of Docker build completion.
  * Improved deployment SHA resolution for better consistency across rollbacks, releases, and manual deployments.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -137,14 +137,14 @@ jobs:
                   detect_interval=5
                   run_info=""
                   while [ $detect_elapsed -lt $detect_max ]; do
-                    run_info=$(gh run list \
-                      --repo "${{ github.repository }}" \
-                      --workflow=docker-publish.yml \
-                      --json headSha,status,conclusion,databaseId \
-                      --limit 10 \
-                      | jq -r --arg sha "$commit_sha" \
-                        '.[] | select(.headSha == $sha) | "\(.databaseId) \(.status) \(.conclusion)"' \
-                      | head -1)
+                    # Look the build up by exact head_sha, not from the N most
+                    # recent runs: a release can be published for an older tag
+                    # whose docker-publish run has since fallen out of the recent
+                    # window, which a capped `gh run list` would miss (#1398).
+                    run_info=$(gh api \
+                      "repos/${{ github.repository }}/actions/workflows/docker-publish.yml/runs?head_sha=${commit_sha}&per_page=1" \
+                      --jq '.workflow_runs[0] | select(. != null) | "\(.id) \(.status) \(.conclusion)"' \
+                      2>/dev/null || true)
                     [ -n "$run_info" ] && break
                     echo "  No docker-publish run yet, waiting ${detect_interval}s... (${detect_elapsed}s elapsed)"
                     sleep $detect_interval
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -13,10 +13,8 @@ on:
                 required: false
                 default: ''
                 type: string
-    workflow_run:
-        workflows: ["Build & Push Docker Images"]
-        types: [completed]
-        branches: [main]
+    release:
+        types: [published]
 
 permissions:
     contents: read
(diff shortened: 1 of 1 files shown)


LucasSantana-Dev:
Fixed in ffe13e15 (@cubic-dev-ai fyi): `resolve_sha` now rejects anything but a 7–40 char lowercase hex SHA before exporting, and the four steps that still inlined the output (docker_wait, Sentry create/finalize, homelab wait) now read it from `env` — matching the webhook/validate steps and the workflow's pre-existing injection-safe pattern. Used 7–40 rather than strict 40 so short rollback SHAs still work.

LucasSantana-Dev:
@cubic-dev-ai P2 fixed in 0b4902a2: `docker_wait` now looks the build up by exact `head_sha` via the Actions API (`workflows/docker-publish.yml/runs?head_sha=…`) instead of scanning the 10 most recent runs — so a release published for an older tag still finds its build regardless of how many builds ran since. Same downstream id/status/conclusion handling and not-yet-queued retry.

PR #1489, opened 2026-06-17
fix(help): split large command categories across embed fields

### LINKED ISSUE #1487: fix(help): /help crashes — embed field value exceeds Discord 1024-char limit
## Summary
`/help` throws and shows a generic error. `createHelpEmbed` packs every command in a category into a single Discord embed **field value**, with no length guard. Discord's hard limit is 1024 chars per field value; the Music category (90+ commands) blows past it, so discord.js throws a `RangeError` during serialization, which is swallowed into a generic message.

## Evidence
User-facing error (live test, screenshot):
```
Error
An error occurred while displaying the help commands.
```

Root cause — `packages/bot/src/functions/general/commands/help.ts:61-68`:
```ts
for (const { key, label } of categories) {
    if (categoryCommands[key].length > 0) {
        embed.addFields({
            name: `${label} (${categoryCommands[key].length})`,
            value: `​\n${categoryCommands[key].join('\n')}`, // line 65 ← no 1024-char cap
            inline: false,
        })
    }
}
```
The real error is swallowed at `help.ts:101-113` and only logged via `errorLog({ message: 'Help command error:', error })` (line ~92). The footer reports ~160 total commands, so Music alone (~90) far exceeds 1024 chars.

Discord limits are codified in `packages/shared/src/services/embedValidation.ts` (field value ≤1024, ≤25 fields/embed). `queueEmbed.ts` truncates correctly; `help.ts` does not.

## Fix (recommended)
Chunk each category's command list into multiple fields, each ≤1024 chars, capping total fields ≤25 (split "Music" across several "Music (cont.)" fields). Truncation-with-ellipsis is the minimal alternative but drops most Music commands.

## Repro
`/help` in Discord → "An error occurred while displaying the help commands." Reproduces on clean `main` (unconditional; grows with command count).

LucasSantana-Dev:
## Problem
`/help` threw and showed "An error occurred while displaying the help commands." `createHelpEmbed` packed every command in a category into a single embed **field value** with no length guard. The Music category (90+ commands) exceeded Discord's 1024-char field limit, so discord.js threw a `RangeError` during serialization, swallowed into the generic message.

## Fix
Chunk each category's command list into multiple fields, each ≤1024 chars, capped at Discord's 25-field maximum. Continuation fields are labelled `<Category> (cont.)`. No commands are dropped (unless the total would exceed 25 fields, which is logged-by-design via the cap).

## Test
tsc passes (pre-commit type:check green). The chunker never splits a single command line across fields and respects the field-value budget.

Closes #1487

---
## Summary by cubic
Split long /help categories across multiple embed fields and paginate across embeds to respect Discord limits and prevent crashes. Fixes #1487 by avoiding `discord.js` RangeError and ensuring all commands are shown.

- **Bug Fixes**
  - Chunk each category’s commands into field values ≤1024 chars and page fields across embeds to stay under 25 fields and the 6000-char cap.
  - Label continuation fields as "<Category> (cont.)", never split a command line, send each page as a follow-up message with page titles, and use `\u200B` instead of a literal zero-width space.

<sup>Written for commit 342482c1e914cf59f3aa5e9c6b66caefd2c2a233. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/1489?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/bot/src/functions/general/commands/help.ts
+++ b/packages/bot/src/functions/general/commands/help.ts
@@ -66,7 +66,7 @@ function createHelpEmbeds(
                 index === 0
                     ? `${label} (${commands.length})`
                     : `${label} (cont.)`
-            fields.push({ name, value: `​\n${chunk}` })
+            fields.push({ name, value: `\u200B\n${chunk}` })
         })
     }
 
@@ -125,7 +125,7 @@ function createHelpEmbeds(
 
 const MAX_FIELD_VALUE = 1024
 const MAX_EMBED_FIELDS = 25
-// Leave headroom for the `​\n` prefix added to each field value.
+// Leave headroom for the zero-width-space + newline prefix on each field value.
 const FIELD_VALUE_BUDGET = MAX_FIELD_VALUE - 8
 // Discord caps a message at 6000 chars across all embeds; keep headroom for
 // the first page's title, description and footer.
--- a/packages/bot/src/functions/general/commands/help.ts
+++ b/packages/bot/src/functions/general/commands/help.ts
@@ -58,17 +58,60 @@ function createHelpEmbed(
             iconURL: interaction.user.displayAvatarURL(),
         })
 
+    let fieldCount = 0
     for (const { key, label } of categories) {
-        if (categoryCommands[key].length > 0) {
-            embed.addFields({
-                name: `${label} (${categoryCommands[key].length})`,
-                value: `\u200B\n${categoryCommands[key].join('\n')}`,
-                inline: false,
-            })
+        const commands = categoryCommands[key]
+        if (commands.length === 0) {
+            continue
         }
+
+        // Discord caps embed field values at 1024 chars and embeds at 25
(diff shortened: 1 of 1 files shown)


LucasSantana-Dev:
Addressed: the original fix only guarded the per-field (1024) and field-count (25) limits. Discord also caps a message at 6000 chars across all embeds, which ~160 commands exceed — so the embed send would still have thrown. Reworked to paginate the fields into multiple embeds and send each as its own message (the reply helper routes follow-on calls to `followUp`), keeping every message under 6000 and dropping no commands.

PR #1543, opened 2026-06-23
fix(backend): harden role + reaction-role write-path error handling

### LINKED ISSUE #1539: bug: GuildService role write paths swallow errors via REST fallback (wrong 404 mapping, duplicate-create, broken idempotent delete)
## Root cause
`GuildService` role write methods (added in #1524) use the shape `if (client) { try { …bot-client op…; return } catch { debugLog } }` then fall through to a REST call. This treats **every** bot-client error — including legitimate "role not found" and post-mutation failures — as a signal to retry via REST, producing three distinct bugs (flagged by CodeRabbit on the v2.21.0 delta).

## Bugs (all in `packages/backend/src/services/GuildService.ts`)

1. **updateGuildRole — not-found swallowed → wrong HTTP status** (~L966)
   `if (!role) throw new Error('Role not found')` is *inside* the try, so it's caught at the `catch` (~L991), debugLog-ged, and discarded. Execution falls into the REST PATCH, which on a missing role throws a generic `Discord API error: …` instead. The roles route maps `'Role not found'` → 404, so with the bot client connected it returns **badRequest instead of 404**.

2. **createGuildRole — duplicate role on transient failure** (~L864)
   If `guild.roles.create(...)` succeeds on Discord but the call rejects afterward (transient network / post-create mapping error), control falls through to the REST `POST …/roles` branch and a **second role with the same name** is created. The REST path should only be a fallback for client *unavailability* (null / not in guild), not for operational failures of the attempt.

3. **deleteGuildRole — non-idempotent delete** (~L1052)
   When `guild.roles.fetch(roleId)` resolves `null` (already gone), the `if (role)` block is skipped, there is no `return`, and execution drops into the REST DELETE which 404s and **throws** — even though the bot client already confirmed the role is absent. Breaks idempotent delete and blocks the route's notFound mapping.

## Fix direction
Gate the REST fallback on "client cannot serve this guild" (client null / not in guild), NOT on operational failures. Let not-found be a definitive result in the client path (return for delete idempotency, or re-throw `'Role not found'` past the fallback so the route maps 404). 

## Evidence
Confirmed by reading the three regions on `main` @ b05a9cfe. Pre-existing from #1524 (shipped in v2.21.0). Not a security/data-loss-on-happy-path issue → not release-blocking, but should be the next role-management fix. Related: #1527 (add GuildService tests — the tests added there should cover these paths).

### LINKED ISSUE #1542: bug: ReactionRolesService swallows DB errors (list/delete) + orphan message on partial create
Found by /backend-slop-audit on the v2.21.0 delta (`packages/shared/src/services/ReactionRolesService/index.ts`). Distinct from the validation hardening in #1540.

## Findings

1. **Swallowed DB error → misleading empty list (Major, #1 swallowed-error)** — `listReactionRoleMessages` (L326-342) `catch { return [] }`. On a DB outage the dashboard shows "no reaction roles" — indistinguishable from a guild that genuinely has none. Let it throw (or return a typed error) so the route can surface 5xx.

2. **Swallowed DB error → conflated delete result (Major, #1/#8)** — `deleteReactionRoleMessage` (L292-323) returns `false` for BOTH "not found / wrong guild" (legit) AND a DB error (failure). Route (roles.ts:78) maps `false`→404, so a DB failure returns a misleading 404. Distinguish not-found from failure (return false only for not-found; let DB errors propagate).

3. **Orphan Discord message on partial create (Major, #20 missing compensation)** — bot-client `createReactionRoleMessage` (L104-139, wired at `packages/bot/src/functions/general/handlers/reactionroleHandlers.ts:69`): if `prisma.reactionRoleMessage.create` throws AFTER `channel.send` succeeds, the catch logs+rethrows but the sent message stays — a live message whose buttons resolve to 'no longer valid'. The dashboard variant (`createReactionRoleMessageFromDashboard`, L263-274) already compensates by deleting the Discord message; mirror that here (best-effort `message.delete()` in the catch before rethrow).

## Evidence
Read on `main` @ c199346b. Happy-path tests pass; these are failure-mode gaps. Shipped in v2.21.0 via #1532. Pairs with #1539 (the GuildService REST-fallback twin of the same 'swallow + fall through' anti-pattern).

LucasSantana-Dev:
## Summary
Fixes the backend write-path error-handling anti-patterns surfaced by CodeRabbit + `/backend-slop-audit` on the v2.21.0 delta. Two clusters, both test-first (XP), each spec+quality reviewed.

### #1539 — GuildService role writes (`packages/backend`)
The `if (client) { try { … } catch { debugLog } }` → REST fallback treated **every** bot-client error as a retry signal. Extracted `getServableGuild()` (catches only client-*unavailability*); the bot-client path is now authoritative — operational errors propagate, REST runs only when the client is null / not in the guild. Fixes:
- `updateGuildRole` / `deleteGuildRole` missing role → throws `'Role not found'` → **404** (was 400 / 500)
- `createGuildRole` transient failure → no fall-through → no **duplicate role**
- +8 unit tests (incl. a regression guard that the REST path still runs when the client is null)

### #1542 — ReactionRolesService (`packages/shared`)
- `listReactionRoleMessages` propagates DB errors instead of returning `[]` (no more masked outage)
- `deleteReactionRoleMessage` returns `false` only for not-found / wrong-guild; DB errors propagate
- bot-client `createReactionRoleMessage` deletes the sent Discord message if the Prisma write fails (no orphan message with dead buttons — mirrors the dashboard variant)

## Tests
- backend: GuildService 46 + roles integration 16 — green
- shared: ReactionRolesService spec **60** — green
- All new tests verified RED before the fix, GREEN after (TDD).

Closes #1539
Closes #1542

---
## Summary by cubic
Tightened error handling for role and reaction-role writes to stop masking real failures and prevent duplicates. Bot-client errors now bubble up; REST runs only when the client is unavailable. Addresses #1539 and #1542.

- **Bug Fixes**
  - `packages/backend` (`GuildService` roles)
    - Added `getServableGuild()`; bot client is the source of truth. REST fallback only when the client is null or not in the guild.
    - Update/delete of a missing role now throws "Role not found" → 404.
    - No fall-through after client errors on create, preventing duplicate roles.
  - `packages/shared` (`ReactionRolesService`)
    - `listReactionRoleMessages` now propagates DB errors (no silent empty array).
    - `deleteReactionRoleMessage` returns false only for not-found/wrong guild; DB errors propagate, except Prisma `P2025` (concurrent delete) which returns false.
    - `createReactionRoleMessage` deletes the Discord message if the Prisma write fails to avoid orphaned buttons.

<sup>Written for commit 03b6cd155877e5dc52f1607fd6a2264c38d9061b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/1543?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/backend/src/services/GuildService.ts
+++ b/packages/backend/src/services/GuildService.ts
@@ -105,6 +105,19 @@ class GuildService {
         return getClient()
     }
 
+    private async getServableGuild(guildId: string): Promise<Guild | null> {
+        const client = this.getBotClient()
+        if (!client) return null
+        try {
+            return (
+                client.guilds.cache.get(guildId) ??
+                (await client.guilds.fetch(guildId))
+            )
+        } catch {
+            return null
+        }
+    }
+
     clearBotGuildCache(): void {
         this.botGuildIdsCache = null
         this.botGuildIdsInFlight = null
@@ -865,39 +878,29 @@ class GuildService {
         guildId: string,
         data: RoleUpsertData,
     ): Promise<GuildRoleManage> {
-        const client = this.getBotClient()
+        const guild = await this.getServableGuild(guildId)
 
-        if (client) {
-            try {
-                const guild =
-                    client.guilds.cache.get(guildId) ??
-                    (await client.guilds.fetch(guildId))
-                const role = await guild.roles.create({
-                    name: data.name,
-                    color: data.color,
-                    hoist: data.hoist,
-                    mentionable: data.mentionable,
-                    permissions:
-                        data.permissions !== undefined
-                            ? BigInt(data.permissions)
-                            : undefined,
-                    reason: 'Created via dashboard',
-                })
-                return {
-                    id: role.id,
-                    name: role.name,
-                    color: role.color ?? 0,
-                    hoist: role.hoist ?? false,
-                    mentionable: role.mentionable ?? false,
-                    permissions: role.permissions.bitfield.toString(),
-                    position: role.position ?? 0,
-                    managed: role.managed ?? false,
-                }
-            } catch (error) {
-                debugLog({
-                    message: 'Failed to create guild role via bot client',
-                    error,
-                })
+        if (guild) {
+            const role = await guild.roles.create({
+                name: data.name,
+                color: data.color,
+                hoist: data.hoist,
+                mentionable: data.mentionable,
+                permissions:
+                    data.permissions !== undefined
+                        ? BigInt(data.permissions)
+                        : undefined,
+                reason: 'Created via dashboard',
+            })
+            return {
+                id: role.id,
+                name: role.name,
+                color: role.color ?? 0,
+                hoist: role.hoist ?? false,
+                mentionable: role.mentionable ?? false,
+                permissions: role.permissions.bitfield.toString(),
+                position: role.position ?? 0,
+                managed: role.managed ?? false,
             }
         }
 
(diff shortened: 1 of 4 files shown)


LucasSantana-Dev:
Addressed cubic review:
- **delete P2025** (`index.ts:314`): fixed in `03b6cd15` — a concurrent delete that races to `P2025` now returns `false` (not-found contract preserved), with a regression test.
- **getServableGuild fetch errors** (`GuildService.ts:118`): keeping as designed. Returning `null` on a guild-fetch failure is the intended *availability* signal — control then takes the REST path, whose result (success **or** error) fully propagates. The PR's core fix is that the *write operation's* errors (role create/edit/delete) no longer fall through to REST; the availability probe falling back is not a swallow.

PR #1609, opened 2026-06-25
feat(bot): add weekly community digest service

LucasSantana-Dev:
Implements weekly digest feature that posts every Monday 12:00 UTC to the configured digest channel:

## Features

- **Top-reacted messages**: Fetches top 3 messages from forum channel (last 7 days), sorted by reaction count
- **Member count delta**: Tracks week-over-week community growth via WeeklyDigestSnapshot model
- **Upcoming events**: Lists guild scheduled events in the next 7 days
- **Automatic scheduling**: Uses hourly tick pattern to check for Monday 12:00 UTC window

## Setup

- Add Prisma model: `WeeklyDigestSnapshot` for member count tracking
- Create migration: Adds `weekly_digest_snapshots` table
- Create service: `WeeklyDigestService` with interval-based scheduling
- Register in initializer: Service started on bot ready, stopped on shutdown

## Configuration

The service is enabled via environment variables:
- `DIGEST_CHANNEL_ID` — where digests are posted
- `FORUM_CHANNEL_ID` — where top messages are fetched from

If either is unset, the service remains idle.

## Error Handling

All digest operations wrapped in try/catch. Failures are logged but non-fatal.
Single-flight mechanism prevents duplicate sends within 30 minutes.

## Summary by CodeRabbit

* **New Features**
  * Added a weekly digest that posts to configured channels with member/community stats, top posts, and upcoming events.
  * The digest avoids duplicate sends by tracking a weekly snapshot of guild member counts.
* **Bug Fixes**
  * Improved bot startup/shutdown so the weekly digest service starts and stops reliably.
  * Enhanced shutdown resilience so teardown continues even if the weekly digest stop encounters an error.
* **Tests**
  * Added comprehensive coverage for weekly digest scheduling, sending, cooldown behavior, and shutdown cleanup.


LucasSantana-Dev:
### Code review

Found 4 issues:

1. `getTopReactedMessages()` throws on the intended channel type: the config check accepts `ChannelType.GuildForum`, but the channel is then treated as a `TextChannel` and `.messages.fetch()` is called on it — forum channels have `.threads`, not `.messages`, so a configured forum channel (env is `CRIATIVARIA_FORUM_CHANNEL_ID`; PR description says forum) fails at runtime and top-reacted messages never work. Branch on channel type and iterate threads (e.g. `fetchActive()` + starter/thread messages) for forums.

https://github.com/LucasSantana-Dev/Lucky/blob/d51d06275cb97333ee10c9cf3184b01e0a1f0262/packages/bot/src/services/WeeklyDigestService.ts#L279-L286

2. Snapshot persisted before the digest send: `prisma.weeklyDigestSnapshot.create()` runs before `digestChannel.send()`, and the catch path doesn't undo it — a failed send (rate limit, permissions, network) advances the weekly baseline, the idempotency check suppresses retry, and that week's digest is silently lost. Persist (or update) the snapshot only after the send succeeds. Same ordering issue was raised on #1608 and #1613.

https://github.com/LucasSantana-Dev/Lucky/blob/d51d06275cb97333ee10c9cf3184b01e0a1f0262/packages/bot/src/services/WeeklyDigestService.ts#L253-L263

3. Leftover debug `console.log` statements bypass the shared logger: `console.log('Tick early return', ...)` and `console.log('Not Monday 12 UTC', ...)` fire every hourly tick in prod stdout, outside the `debugLog`/`infoLog` pipeline this same file already uses everywhere else.

https://github.com/LucasSantana-Dev/Lucky/blob/d51d06275cb97333ee10c9cf3184b01e0a1f0262/packages/bot/src/services/WeeklyDigestService.ts#L105-L124

4. Tick-interval env parsing accepts partial numerics: `Number.parseInt(raw, 10)` turns `"1h"` into `1` (a 1 ms tick — hot loop) and the finite/positive guards don't catch it. The shared `parseIntEnv` helper in `packages/shared/src/utils/env.ts` already does full-string validation (`/^[+-]?\d+$/`) — use it instead of the local reimplementation.

https://github.com/LucasSantana-Dev/Lucky/blob/d51d06275cb97333ee10c9cf3184b01e0a1f0262/packages/bot/src/services/WeeklyDigestService.ts#L38-L40

Checked for bugs, CLAUDE.md compliance, git-history regressions, prior-PR feedback, and in-code guidance. Notable near-misses (verified, below report threshold): `WeeklyDigestSnapshot` has no week-scoped unique key (check-then-create race is restart-window-only for a single-instance bot), and the new `criativariaLiveNotificationService` mock isn't exercised in the initializer spec's `failingStops` matrix.

LucasSantana-Dev:
Review findings addressed in d10485ab:

1. **GuildForum support** — `getTopReactedMessages` now branches on `ChannelType.GuildForum` and collects active-thread starter messages via `threads.fetchActive()` + `fetchStarterMessage()`; text channels keep the last-100 fetch. Regression test added.
2. **Send-before-persist** — `weeklyDigestSnapshot.create()` moved after `digestChannel.send()`; failed send no longer advances the weekly baseline. Test now asserts `create` is NOT called when send throws.
3. **console.log leftovers** — both replaced with `debugLog` (structured pipeline).
4. **Env parsing** — local `parsePositiveIntEnv` deleted; uses shared `parseIntEnv` (full-string validation, `{ min: 1 }`). Added `"1h"` rejection test.

Bonus fix found while running the suite: `uses previous snapshot for member delta` was a time-bomb — `postedAt: new Date()` (real wall clock) drifted out of the mocked week (clock pinned to 2026-06-22), tripping the idempotency skip; it fails on clean HEAD today. Now anchored to the mocked clock minus 7 days.

Suite: 65/65 passing (26 digest + 39 initializer).

PR #1816, opened 2026-07-12
perf(docker): compile @discordjs/opus once, reuse in prod stage

LucasSantana-Dev:
## Summary
Eliminates double compilation of @discordjs/opus in Docker builds.

- **Before:** Compiled in build stage (line 71), then again in deps-production stage (line 121)
- **After:** Copy pre-built node_modules from build stage, prune devDeps in-place

## Verification required
UNVERIFIED — docker daemon not available in this environment. 

To verify:
```bash
docker build --target production-bot -t lucky-opus-test .
docker run --rm lucky-opus-test node -e "require('@discordjs/opus'); console.log('opus OK')"
docker images lucky-opus-test --format '{{.Size}}'
```

Should print 'opus OK' and show image size ≤ current main. Confirm opus compiles once only in build log.

## Technical notes
- Reuses pattern already used for @prisma/engines at line 150
- npm prune --omit=dev removes devDeps while preserving pre-compiled .node binary
- Reduces build time by ~30-45 minutes per image (opus source compile dominates)

---
## Summary by cubic
Compile `@discordjs/opus` once during Docker builds by reusing the compiled `node_modules` in the production stage, cutting build time by ~30–45 minutes. CI now tags and loads the bot image and verifies native modules at runtime to catch regressions.

- **Refactors**
  - Reuse pre-built `node_modules` from the build stage (root and workspaces) and prune in place with `npm prune --omit=dev --legacy-peer-deps` to keep the compiled `.node`.

- **Bug Fixes**
  - CI tags `lucky-bot:ci`, loads it for the bot only, and requires `@discordjs/opus` to encode a frame to validate native addons.

<sup>Written for commit ba4c2211989a537db1bdadc39edfc559c2c0c0aa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/1816?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved production container dependency handling by reusing build artifacts and pruning to production-only packages.
  * Reduced differences between build-time and production environments for more consistent deployments.
* **CI / Quality**
  * Added deterministic image tagging during CI and adjusted build behavior per service.
  * Added a post-build check for the bot image to ensure native modules load correctly.


LucasSantana-Dev:
## Re: the two CodeRabbit 🟠 Major findings — both tested, both refuted

**`Dockerfile:127` — "npm prune only prunes the root install, add `--workspaces --include-workspace-root`"**

Tested directly on a minimal npm-workspace fixture (`node:24-alpine`, npm semantics are arch-independent), with a control case to prove the instrument actually measures something:

| command | devDep `is-odd` | prodDep `ms` |
|---|---|---|
| after `npm install` (control) | 1 | 1 |
| `npm prune --omit=dev` ← what this PR runs | **0** | 1 |
| `npm prune --omit=dev --workspaces --include-workspace-root` ← suggested fix | 0 | 1 |

Plain `npm prune --omit=dev` **already removes** the workspace devDependency. The suggested flags are a no-op here. Reason: workspaces hoist devDeps to the **root** `node_modules`, so a root-level prune reaches them.

Note this is exactly what the sibling finding at `Dockerfile:125` asserts ("npm workspaces hoist by default"). The two findings are mutually inconsistent — devDeps cannot both hoist to root *and* remain in nested `packages/*/node_modules`.

**`Dockerfile:125` — "`/app/packages/*/node_modules` isn't guaranteed to exist, COPY can fail the build"**

The build passes, so the directories exist. And a missing `COPY` source is a **loud** failure — it fails the image build, which `Build — Docker images` gates on. It cannot ship silently.

**Residual risk (acknowledged, not dismissed):** if a version conflict ever forces a devDep to nest under `packages/*/node_modules` instead of hoisting, root prune would miss it. Consequence would be image bloat, not breakage. Not observed today.

**What actually needed proving on this PR** was not the prune — it was that the copied `.node` binary loads at runtime. `docker build` never require()s anything, which is how #1735 shipped a Prisma engine that crash-looped 262×. This PR now carries a gate that loads the real `production-bot` image and exercises the addon:

```
opus OK — encoded 3 bytes
```

(3 bytes = Opus compressing a silent 3840-byte frame — expected.) The gate was verified to fail on a missing module (exit 1), an empty frame (exit 1), and a missing image (exit 125), so a green result means something.

PR #1874, opened 2026-07-24
fix(frontend): stop hero grid columns overflowing on narrow viewports

reag-dev:
## Description

Both Hero grid items (headline column and RepoCard column) lacked min-w-0, so CSS Grid's default min-width:auto kept them from shrinking below their content. The RepoCard's git-clone <code> block used `truncate` (white-space:nowrap) inside a flex row without min-w-0, giving it a ~500px intrinsic minimum that inflated the shared grid track below the lg breakpoint and pushed the headline column past the viewport width.

Adds min-w-0 to both grid item wrappers and min-w-0 flex-1 to the clone <code> element so it actually truncates. Verified with headless Playwright across 320-1280px viewports: no horizontal overflow left.

## Checklist
- [ X ] Tests pass locally
- [ X ] CHANGELOG.md updated (if user-facing)
- [ X ] TypeScript builds cleanly

Closes issue #1825 

---
## Summary by cubic
Fixes #1825 by preventing the landing hero grid from overflowing on small screens. Lets both columns and the clone command shrink, removing horizontal scroll.

- **Bug Fixes**
  - Add `min-w-0` to both hero grid item wrappers.
  - Make the clone `<code>` element `min-w-0 flex-1` so `truncate` works inside the flex row.

<sup>Written for commit 92daa49fd63697acf630a8a242246ace247a15f2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/1874?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved layout behavior in the landing page hero and repository card to better support shrinking/truncation.
  * Long “git clone” commands now truncate correctly instead of overflowing their container.

PR #1873, opened 2026-07-24
ci(deps): reactivate dependabot with grouped minor/patch updates

LucasSantana-Dev:
Dependabot: npm, docker, and github-actions ecosystems on weekly interval with open-pull-requests-limit 5 and minor/patch grouping (majors stay separate). Stale-PR workflow skipped: repo already has .github/workflows/stale.yml.

---
## Summary by cubic
Reactivates Dependabot with grouped weekly minor/patch updates for `npm`, `docker` (root and `/deploy`), and `github-actions` (max 5 open PRs); majors stay separate. Adds cooldowns (`default-days: 3` for all; `semver-major-days: 7` for `npm` only), excludes YouTube-adjacent `npm` packages to keep smoke tests effective, removes Renovate (`.renovaterc.json`, `.github/workflows/renovate-health.yml`), and adds an ADR; stale PR workflow stays as-is.

<sup>Written for commit 4db9e07ae5cf6e3d2929d7e24d1a2345ae13c562. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/1873?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Chores**
  * Reactivated automated dependency updates with weekly checks and controlled PR volume.
  * Streamlined updates by grouping minor/patch changes while keeping select video-related packages as standalone updates.
  * Improved coverage for container builds, including the deployment directory.
  * Removed the prior scheduled “Renovate health” checking job and retired the old Renovate configuration.
* **Documentation**
  * Added an ADR explaining the reactivation approach and how it replaces earlier decisions.

PR #1876, opened 2026-07-26
fix(ci): make the security gate passable and clear production advisories

LucasSantana-Dev:
Splits out of #1875. That PR's Dockerfile fix is blocked by the required `Security` check, which fails on `main` and on every other open PR (#1873, #1874), not on anything #1875 changed.

## What this clears

`npm audit --omit=dev` production high/critical: **9 -> 6**.

**`find-my-way` -> `>=9.7.0` override.** The HTTP/2 DDoS advisory covers `<=9.6.0`; 9.7.0 is published. It reaches us through `@prisma/dev` <- `prisma`, and `prisma` is already at its latest (7.9.0), so an override is the only route that does not wait on upstream. Clears three reported findings: `find-my-way`, `@prisma/dev`, `prisma`.

**`brace-expansion` override `>=5.0.6` -> `>=5.0.8`.** The advisory range is `<=5.0.7`, so the previous pin targeted a version that was itself vulnerable. That is why it never took effect.

**Lockfile resync** (`eslint` 10.8.0, `@eslint/config-helpers` 0.7.0). `npm ci --legacy-peer-deps` on npm 11.16.0 currently rejects the lock repo-wide:

```
Invalid: lock file's eslint@10.7.0 does not satisfy eslint@10.8.0
```

Nothing here asks for 10.8.0; it was published after the last refresh while root carries `^10.4.0`. This is what breaks `Build — shared`, and what makes `compressed-size` fail on its base-branch install. Same class as #1809.

## What this deliberately does not fix

Each needs a major-version decision that should not be made silently:

| finding | fix | why not here |
|---|---|---|
| `brace-expansion`, `minimatch`, `glob`, `rimraf` | force `brace-expansion@5.0.8` tree-wide | old `minimatch` copies request `^1.1.7` and `^2.0.2`, so a range override cannot replace them. Forcing it substitutes a major inside a glob-matching library used by `glob`, `rimraf`, `test-exclude`, `@swc/cli` |
| `react-router` | 8.3.0 | `react-router-dom`'s 7 line ends at 7.18.1 and pulls `react-router` 7.x. Needs the v8 upgrade with route-API review |
| `@discordjs/node-pre-gyp` | `@discordjs/opus@0.2.1` | semver-major, and that package already needs musl prebuild special-casing in the Dockerfile |

**`Security` stays red until those are decided**, so this does not by itself unblock #1875. It does remove three real production advisories and repair `npm ci`.

## Note on approach

`npm audit fix --package-lock-only` was tried first and rejected: it rewrote 499 lock lines, raised production high/critical from 9 to 12, and introduced a **critical `tar`** finding. Reverted. These are targeted overrides instead, each verified against its advisory range.

---
## Summary by cubic
Reduces production security findings and stabilizes CI by overriding vulnerable deps, resyncing the lockfile, switching to a production-only audit gate, and bumping frontend packages to clear additional advisories. `npm ci` no longer fails on a lock mismatch.

- **Dependencies**
  - Override `find-my-way` to `>=9.7.0` (HTTP/2 DDoS on `<=9.6.0`); resolves the chain via `@prisma/dev` → `prisma`.
  - Override `brace-expansion` to `>=5.0.8`.
  - Bump `react-router-dom` to `^7.18.1` to clear four advisories; only the RSC-only `react-router` advisory stays accepted until the v8 migration (now reported transitively via `react-router`).
  - Bump `postcss` to `^8.5.23` to clear GHSA-r28c-9q8g-f849.
  - Resync lockfile to `eslint@10.8.0` and `@eslint/config-helpers@0.7.0`.

- **CI**
  - Replace `npm audit --audit-level=high` with `scripts/audit-gate.mjs` (audits `--omit=dev`, blocks any unaccepted high/critical).
  - Acceptance is pinned to advisory IDs (or marked TRANSITIVE), enforces non-empty reason/until, and fails if any accepted advisory stops being reported.

<sup>Written for commit 9df42f8f58a9cd6807bea40a88e659c16009588a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/1876?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Security**
  * Strengthened dependency vulnerability checks for production packages.
  * Added enforcement for secure versions of affected dependencies.
  * High- and critical-severity findings now block releases unless explicitly reviewed and accepted.
  * Added validation to remove outdated vulnerability exceptions when they are no longer reported.


LucasSantana-Dev:
Addressed in `ae114485`. Two of the three were right and are fixed.

**Allowlist individual advisories, not package names** — agreed, and this was the important one. Keying on package name meant a newly disclosed advisory in `react-router` would have passed the gate *and* kept the acceptance looking fresh, so the anti-rot check would have stayed quiet about it. `ACCEPTED` entries now pin exact npm advisory ids.

Packages npm reports only because a dependency is vulnerable (`rimraf`, `glob`, `minimatch`, `@discordjs/node-pre-gyp` — their `via` holds package names, not advisory objects) carry a `TRANSITIVE` marker instead, since there is no advisory id to pin. If one of those ever gains a direct advisory, the gate blocks so it gets reviewed rather than inherited.

**Enforce required acceptance metadata** — agreed. `reason` and `until` must now be non-empty strings, and `advisories` must be `TRANSITIVE` or a non-empty map, validated before the audit runs.

I did not take the committable suggestion verbatim, as it wrapped the package keys in backticks (`'`@discordjs/node-pre-gyp`'`), which would have stopped them matching the npm report.

While implementing this I found a bug in my own first pass: a package that was *blocking* also appeared in the stale list as "no longer reported at all", which buried the real reason under a wrong one. Blocking packages are now excluded from the stale calculation.

Verified all five paths rather than assuming:

| scenario | result |
|---|---|
| baseline | passes, exit 0 |
| pinned advisory removed from list (simulates a new CVE) | blocks: `react-router — advisory not in the accepted list: 1124276 GHSA-chx6-hx7r-mcp5` |
| whole package unaccepted | blocks: `rimraf — not accepted` |
| entry missing `until` | config error before the audit runs |
| pinned advisory no longer reported | flagged stale |

LucasSantana-Dev:
Good catch, agreed and fixed in `9df42f8f`.

You're right that accepting all five react-router advisories masked four that had a compatible fix inside the 7.x line. Bumped `react-router-dom` `^7.15.1` -> `^7.18.1`, which clears:

- GHSA-wrjc-x8rr-h8h6, open redirect via backslash in `<Link>`/`useNavigate`
- GHSA-h8fp-f39c-q6mh, RSCErrorHandler missing protocol validation (XSS)
- GHSA-337j-9hxr-rhxg, constructor injection via `deserializeErrors()`
- GHSA-chx6-hx7r-mcp5, DoS via inefficient route matching

Only GHSA-qwww-vcr4-c8h2 stays accepted: fixed in 8.3.0 only, and RSC-mode specific while this app is a Vite SPA. The v8 migration is tracked in #1878, and the gate's anti-rot check will force that entry to be deleted when it lands.

Two things that fell out of it:

- Re-resolving the lock surfaced `postcss` GHSA-r28c-9q8g-f849 as a production high. The declared range (`^8.5.15`) already allowed the fix, so it's updated rather than accepted.
- `react-router-dom` now reports as a transitive high via `react-router`, so it gets a `TRANSITIVE` entry.

Frontend suite green on the bump: 85 files, 1015 tests.

PR #1881, opened 2026-07-26
fix(docker): treat an empty db password as missing in compose guards

LucasSantana-Dev:
## What

`docker-compose.yml` guarded the Postgres password with `${POSTGRES_PASSWORD?POSTGRES_PASSWORD_required}`. Docker Compose fires that form **only when the variable is absent**. A present-but-empty value passes straight through, so `POSTGRES_PASSWORD=` in `.env` starts Postgres with an empty password and no warning.

Verified against `docker-compose` 5.1.4:

| `.env` state | `${VAR?err}` | `${VAR:?err}` |
|---|---|---|
| unset | errors | errors |
| `POSTGRES_PASSWORD=` (empty) | **passes** | errors |
| real value | passes | passes |

Both interpolation sites now use `:?`, so the guard means what it looks like it means. Confirmed against the real compose file: empty value now fails with `required variable POSTGRES_PASSWORD is missing a value: POSTGRES_PASSWORD_required`, a real value renders the URL normally.

Also documents generating the value in the README setup block. It is interpolated straight into a `postgres://` URL, so it has to be URL-safe, hence `openssl rand -hex 24`. Duplicate keys in `.env` resolve last-wins (verified), so `>>` is safe, and it avoids the `sed -i` GNU/BSD portability trap in a public README.

## Why now

Came out of reviewing #1674, which proposes an **active** 48-zero placeholder in `.env.example`. That would permanently disable this guard for anyone following the documented `cp .env.example .env` flow. The fix there is an empty placeholder, which only works once the guard actually rejects empty values, which is this PR.

This is a pre-existing weakness, not something #1674 introduced.

## Decision record

`decisions/2026-07-26-postgres-password-env-guard.md` covers the alternatives, including the `scripts/setup-env.sh` option and why it is the most likely thing to revisit.

## Not addressed here

Changing `POSTGRES_PASSWORD` in `.env` after first boot does not change the password, because `PGDATA` is initialised once and the volume persists the original. Pre-existing; filed separately.

---
## Summary by cubic
Treats an empty `POSTGRES_PASSWORD` as missing in Docker Compose so dev and staging stacks fail fast instead of starting Postgres with an empty password. Updates the README to generate a URL‑safe password and adds a decision record.

- **Bug Fixes**
  - Switched guards to `${POSTGRES_PASSWORD:?POSTGRES_PASSWORD_required}` in `docker-compose.staging.yml` and `docker-compose.dev.yml` for `DATABASE_URL`, `DIRECT_URL`, and the Postgres service.

- **Migration**
  - Set a non-empty, URL-safe `POSTGRES_PASSWORD` before `docker compose up`, e.g. `echo "POSTGRES_PASSWORD=$(openssl rand -hex 24)" >> .env`

<sup>Written for commit eacd52b72bab3baee3fe1535d7f0ac7ac225c677. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/1881?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LucasSantana-Dev (commented):
## Review: changes-required (self-review pass)

The core `?` → `:?` fix is correct and well-evidenced for `docker-compose.yml`, but the identical weak guard survives untouched in `docker-compose.staging.yml` and `docker-compose.dev.yml`, including the `DIRECT_URL` sites the decision doc says this covers.

**P1 — fix incomplete across compose files:** `docker-compose.staging.yml:46,47,69` and `docker-compose.dev.yml:9,65,66` keep `${POSTGRES_PASSWORD?...}` (6 sites). Concrete failure: `POSTGRES_PASSWORD=` (present but empty) in the staging env, then `scripts/deploy-staging.sh:24` runs Compose; the `?` form errors only on unset, so the empty value passes and Postgres boots with an empty password. Unlike production, where `deploy.sh:550-552` independently hard-fails, `deploy-staging.sh` has no password guard, so the compose guard is the only line of defense on that path. Fix: apply the same one-character `?` → `:?` change to the 6 remaining sites.

Second angle on the same finding: `decisions/2026-07-26-postgres-password-env-guard.md` opens by enumerating "three places" in `docker-compose.yml` including `DIRECT_URL ... # added by #1674` — but `docker-compose.yml` has no `DIRECT_URL` (grep: no match); those lines live in the staging/dev compose files (added by #1830). Intent-vs-implementation mismatch, not just scope.

**P3:** update the decision doc to say the guards were hardened in all three compose files. (The README `echo ... >> .env` duplicate-key trade is disclosed and justified — no action.)

**What's good:** the central `?` vs `:?` claim is empirically verified against docker-compose 5.1.4 with a state table; the `>>` over `sed -i` GNU/BSD portability rationale and the URL-safe `openssl rand -hex 24` justification are both correct and documented; the decision record follows repo convention with alternatives, consequences, and revisit triggers.

Issue #1886, opened 2026-07-26
chore(bot): run the one-off guild-scoped command cleanup


LucasSantana-Dev:
Follow-up to #1885.

Slash-command registration moved from guild-scoped to global. Guild-scoped copies written by earlier versions live on Discord's side until explicitly deleted, and they shadow the global set with whatever definitions the last redeploy wrote, so they have to be removed.

This is deliberately **not** done in the ready handler. Sequencing the REST calls does not sequence Discord's propagation: a newly registered global set is documented as taking up to an hour to become visible, while deleting a guild-scoped set is immediate. Doing both at boot could leave a live server with no commands at all in between. (Caught by cubic on #1887.)

## Steps

1. Deploy the #1887 release so global registration runs.
2. Confirm the global set is actually visible — add the bot to a server it has never been in and check the command picker, or check `GET /applications/{application_id}/commands`.
3. Only then run:

```bash
npm run commands:clear-guild --workspace=packages/bot
```

The script clears each guild independently, so one failure does not abort the rest, and it exits non-zero if any guild was missed. Re-run until it exits 0.

## Exit condition

Every pre-existing guild reports `Cleared guild-scoped commands: <name>`. Until then those guilds keep showing command definitions frozen at the last redeploy.

Worth a CHANGELOG note so a guild that somehow retains guild-scoped commands later (for example one the bot was removed from and re-added to during the window) is not a surprise.

LucasSantana-Dev:
## State after the v2.37.2 deploy

Global registration is live and confirmed on Discord's side, not just in our logs:

```
[2026-07-26T22:27:06.950Z] Global commands registered: 81

GET /applications/{id}/commands -> http 200
global commands on discord: 81
sample: download, afk, birthday, giveaway, help, invite, lastfm, leaderboard
```

No `Guild commands registered:` lines appear at boot any more, so the per-guild loop is gone as intended.

## The cleanup is still needed, and is bigger than expected

Surveyed every guild:

```
guilds still holding guild-scoped commands: 29
```

Each of those 29 holds **80** guild-scoped commands, shadowing the 81 global ones. So they are currently running definitions frozen at whatever the last successful deploy wrote, which was v2.36.1 back on 2026-07-16.

## Deliberately not run yet

[...]

PR #1887, opened 2026-07-26
fix(bot): register slash commands globally, not per cached guild

### LINKED ISSUE #1885: fix(bot): slash commands never register for guilds joined after startup
## Impact

Any server that adds Lucky sees **zero slash commands** until the bot is next restarted. This is the cause of the Top.gg verification rejection (2026-07-26):

> The prefix you've listed on your bot page seems to not match your bot's actual prefix. Please make sure the prefix you list is correct so we are able to test your bot properly. Your prefix is listed as "/" but there are no slash commands deployed to the Top.gg Verification Center.

The reviewer added Lucky to their verification guild, the bot had already been running, and so no commands were ever registered for it. From their side the bot looked non-functional.

It also means every genuine new user has a broken first experience until the next redeploy happens to fix it for them.

## Root cause

`packages/bot/src/handlers/clientHandler/service.ts:92-101` registers commands **per guild**, and only for guilds present in the cache at `ready`:

```ts
for (const guild of client.guilds.cache.values()) {
    await rest.put(
        Routes.applicationGuildCommands(CLIENT_ID, guild.id),
        { body: commandsData },
    )
}
```

There is no registration on join. `handleGuildCreate` (`packages/bot/src/handlers/eventHandler.ts:135`) logs the join, increments `guildJoinsTotal`, records it, and sends the onboarding embed — but never registers commands.

Meanwhile `registerCommands` in `packages/bot/src/handlers/clientHandler/index.ts:8`, which does exactly the correct global registration via `Routes.applicationCommands(clientId)`, has **zero callers**. The right code exists and was never wired up.

## Fix

Register globally at `ready` instead of per guild. Global commands are available in every guild including ones that join later, so there is no join-time race and no O(guilds) REST loop on every boot.

Guild-scoped commands registered by earlier versions persist until explicitly cleared and would otherwise shadow the global set with frozen definitions, so the rollout has to clear them once. Order matters: register global first, then clear guild-scoped, so no guild is left with an empty command list mid-rollout.

## Verify

- Boot the bot, add it to a server it has never been in, confirm slash commands appear without a restart.
- Confirm no duplicate commands in the pre-existing guilds.

Blocks the Top.gg resubmission.

LucasSantana-Dev:
Closes #1885.

## Why

Top.gg rejected the verification on 2026-07-26:

> The prefix you've listed on your bot page seems to not match your bot's actual prefix. Please make sure the prefix you list is correct so we are able to test your bot properly. Your prefix is listed as "/" but there are no slash commands deployed to the Top.gg Verification Center.

The listed prefix was fine. The commands genuinely were not there.

## Root cause

`clientHandler/service.ts` registered commands with `applicationGuildCommands`, looping over `client.guilds.cache` at `ready`. Guilds that added Lucky *while the process was running* were never covered, and `handleGuildCreate` does not register commands either. The reviewer added the bot after boot, so their guild got nothing.

This is not Top.gg-specific: **every** new server has had zero slash commands until the next redeploy happened to include them.

`registerCommands()` in `clientHandler/index.ts` already did the correct global registration and had never been called from anywhere.

## Change

`applicationCommands` registers once and covers every guild including future joins. No join-time race, and no O(guilds) REST loop per boot.

Guild-scoped commands persist server-side until cleared and would otherwise shadow the global set with definitions frozen at the last redeploy, so the ready handler clears them once. The clear runs **after** the global put, never before, so no guild is left without commands mid-rollout. Removing that migration loop is #1886.

## Verification

- New test asserts both puts and their ordering. Confirmed it **fails** against the old per-guild code, so it is a real regression guard rather than a passing decoration.
- Bot suite green: 241 suites, 3176 tests.
- Typecheck and lint clean.

## Rollout note

This needs a **production deploy** before the Top.gg resubmission. Global commands are registered at `ready`, so they will exist before any reviewer joins.

---
## Summary by cubic
Registers slash commands globally to fix missing commands for new servers and Top.gg verification. Moves guild-command cleanup to a one-off script.

- **Bug Fixes**
  - Register once at ready via `Routes.applicationCommands`; remove per-guild writes; new test asserts no guild-scoped calls (#1885).
  - Add cleanup script: `npm run commands:clear-guild` (and `commands:clear-guild:dist`); continues past per-guild failures, reports them, exits non-zero on partial success (#1886/#1887).
  - Paginate `Routes.userGuilds` in the cleanup to cover >200 guilds; add runner tests (TOKEN/CLIENT_ID preflight, failure reporting); exclude `packages/bot/src/scripts/*Cli.ts` from coverage.

- **Migration**
  - Deploy to production before Top.gg resubmission; globals publish at ready.
  - After confirming globals are visible, run `npm run commands:clear-guild` and re-run until it exits 0.

<sup>Written for commit cbbd1e6ce712d7b93dd1a858f9ac5d40d244b47f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/1887?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Slash commands and context menus are now registered globally, making them available across servers.
  * Existing server-specific command copies are automatically cleared during registration.

* **Bug Fixes**
  * Prevented outdated guild-scoped command registrations from persisting after updates.


LucasSantana-Dev:
Both findings were right and are fixed in `010103ef`. The P1 in particular caught a real error in my reasoning, so thanks for it.

**cubic P1 — existing guilds could lose all commands during rollout.** Correct, and my justification was wrong. I had claimed that putting the global registration before the clear meant no guild was ever left empty. That confuses *call ordering* with *propagation ordering*: a newly registered global command set is documented as taking up to an hour to become visible, while deleting a guild-scoped set takes effect immediately. So on the first boot after deploy, the live guilds could have had their instant set removed and nothing to replace it for up to an hour.

The ready handler now only does the global registration, which is all #1885 actually needs — a guild that adds Lucky after boot gets the global set. Nothing guild-scoped is written at boot in either direction.

**CodeRabbit (Major) and cubic P2 — one failing guild aborts the whole cleanup.** Also correct. The `await` sat inside the outer `try`, so a single rejection (bot kicked mid-run, transient 5xx) skipped every remaining guild and left them shadowed, with only a generic error in the outer catch.

The cleanup is now a one-off script, `npm run commands:clear-guild`, run once global registration is confirmed visible. Each guild has its own try/catch, failures are collected and reported by name, and the CLI exits non-zero on partial success so a half-finished migration is not mistaken for a finished one.

Tests updated accordingly:

- the ready handler asserts **no** guild-scoped write happens at boot, in either direction, so neither the original #1885 bug nor the clearing hazard can come back;
- the script asserts it continues past a failing guild and reports which one failed.

Bot suite green: 242 suites, 3178 tests.

LucasSantana-Dev:
Both fixed in `cbbd1e6c`.

**P1 — pagination.** Valid, and the failure mode is the nasty kind: `GET /users/@me/guilds` caps at 200, and reading one page would have skipped every guild beyond it *silently*. A half-run migration would look identical to a finished one, while the skipped guilds kept shadowing the global commands. Lucky is well under 200 today, but growing past it is the entire point of being listed.

Now follows `after` until a short page arrives. Tests cover a single short page (no second request), multi-page with the cursor asserted at each step (a wrong cursor loops on the same 200 forever, so that assertion is the one that matters), and the boundary case of an exactly-full final page followed by an empty one.

**P2 — edge cases.** Added: empty guild list returns `{cleared: [], failed: []}` without issuing any request, and an all-guilds-fail run reports every one of them rather than stopping at the first.

`clearGuildCommands.ts` stays at 100% statements, branches, functions and lines. Bot suite green: 242 suites, 3187 tests.

PR #1889, opened 2026-07-26
fix(frontend): make /invite reach discord and stop asking for admin

### LINKED ISSUE #1888: fix(frontend): /invite link is dead and the docs invite requests Administrator
Two defects in the invite funnel, both found while investigating the Top.gg rejection (#1885). Neither caused that rejection, but both are live and both are risks for the resubmission.

## 1. `/invite` goes to the homepage, not Discord

`https://lucky.lucassantana.tech/invite` is the canonical invite URL. It is used by:

- the Top.gg listing body (`.../invite?utm_source=topgg&utm_medium=directory&utm_campaign=bot-listing`)
- the README "Invite Lucky to Your Server" badge and CTA
- every tracked URL in `.agents/tracking-urls.md`

There is **no `/invite` route**. `App.tsx` defines `/`, `/login`, and a catch-all `<Navigate to='/' replace />`; `vercel.json` rewrites `/:path*` to `/index.html`. So the link serves the SPA, which then bounces to the landing page.

Anyone clicking "Invite" from Top.gg, GitHub, or any tracked campaign lands on the homepage instead of Discord's OAuth consent screen. Every invite CTA in the funnel is broken, and the `utm_source` attribution the tracking doc is built around never records an install.

Verified: `curl -sI https://lucky.lucassantana.tech/invite` returns `200` with the landing page `<title>`, no `Location` header.

## 2. The docs invite requests Administrator

`packages/frontend/src/pages/Docs.tsx:19`

```ts
const BOT_INVITE =
    'https://discord.com/oauth2/authorize?client_id=962198089161134131&scope=bot%20applications.commands&permissions=8'
```

`permissions=8` is **Administrator**.

`packages/frontend/src/pages/Landing.tsx:54` does the opposite, deliberately:

```ts
// Minimal permission set (View Channels, Send Messages, Embed Links, Connect,
// Speak) — matches the /invite command. NOT Administrator: requesting only what
// the bot needs is safer and converts better with server admins.
const BOT_INVITE_PERMISSIONS = '3165184'
```

So the same product ships two invite links with wildly different permission asks, and the Top.gg listing copy explicitly claims:

> ### 🔒 Privacy & Permissions
> Lucky only requests the permissions it actually needs:
> - No admin permission

A reviewer who follows the docs invite sees an Administrator consent screen contradicting the listing. That is a plausible second rejection, and independently it is a bad security default for a self-host-friendly bot.

## Fix

Single source of truth for the invite URL in `packages/frontend/src/lib/discord.ts`, used by both pages, pinned to the minimal permission set. Add a `/invite` redirect so the canonical URL actually reaches Discord.

LucasSantana-Dev:
Closes #1888. Found while investigating the Top.gg rejection (#1885); neither of these caused it, but both are live and both are risks for the resubmission.

## 1. `/invite` never reached Discord

`https://lucky.lucassantana.tech/invite` is the canonical invite URL, used by the Top.gg listing body, the README CTA and every tracked campaign URL in `.agents/tracking-urls.md`.

There was no `/invite` route. `App.tsx` defines `/`, `/login` and a catch-all `<Navigate to='/' replace />`, and `vercel.json` rewrites `/:path*` to `index.html`. So the link served the SPA and bounced to the landing page.

Verified before the fix:

```
$ curl -sI https://lucky.lucassantana.tech/invite
HTTP/2 200
content-type: text/html; charset=utf-8      # landing page, no Location header
```

Every invite CTA in the funnel was a no-op, and the `utm_source` attribution the tracking doc is built on never recorded an install. Fixed with a `redirects` entry in `vercel.json` (Vercel evaluates redirects before rewrites, so the SPA catch-all is unaffected).

## 2. The docs invite requested Administrator

`Docs.tsx` hardcoded `permissions=8`. That is Administrator.

`Landing.tsx` deliberately used `3165184` (View Channels, Send Messages, Embed Links, Connect, Speak) with a comment explaining that Administrator is *not* wanted.

The Top.gg listing states:

> Lucky only requests the permissions it actually needs:
> - No admin permission

So we shipped a link contradicting our own public copy, and a reviewer following the docs invite would have hit an Administrator consent screen. Independently, it is a poor default for a bot that sells itself on minimal permissions.

## Change

One helper in `lib/discord.ts`, used by both pages. The drift was only possible because each page rolled its own URL.

## Tests

`lib/discord.test.ts` asserts the URL never requests Administrator, keeps the `applications.commands` scope (its absence is the shape of #1885), and targets the OAuth endpoint. Verified the guard **fails** when the permission set is put back to `8`:

```
× never requests Administrator
× requests the minimal permission set
Tests  2 failed | 1017 passed
```

Frontend suite green: 86 files, 1019 tests. Typecheck clean.

---
## Summary by cubic
Fixes the invite funnel so `/invite` reaches Discord OAuth and all invite links request minimal permissions (no Administrator). Consolidates the invite URL in one helper and adds tests to prevent regressions.

- **Bug Fixes**
  - Added a `vercel.json` redirect so `/invite` goes to Discord OAuth (bypasses SPA rewrite and keeps UTM).
  - All invite links now use `permissions=3165184` with `applications.commands`; removed `permissions=8` from docs.

- **Refactors**
  - Introduced `getBotInviteUrl` in `packages/frontend/src/lib/discord.ts`, used by Landing and Docs; always returns a URL. Added `packages/frontend/src/lib/discord.test.ts` to assert endpoint, scope, and non-admin permissions.

<sup>Written for commit 6d8b3ff7633104a6f6fa5cf17280a52c1b8e238f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/1889?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LucasSantana-Dev:
Both threads addressed.

**cubic P3 — unreachable disabled-invite path.** Correct, fixed in `ceb45c08`. `DEFAULT_DISCORD_CLIENT_ID` is a non-empty literal, so the `||` fallback always yields an id and `if (!clientId) return ''` could never run. It read like a configuration path without being one. The helper's real contract is that it always returns a URL, so it now says that. (Carried over from the original `Landing.tsx` implementation rather than introduced here, but no reason to keep it.)

**Trivy — "Secret Discord client ID" (MEDIUM).** False positive. A Discord client id is the application id, and it is public by design: it is embedded in every OAuth invite URL, it is visible in the Top.gg listing, and it is already published in `Docs.tsx` and the README invite links. It is not a credential and there is nothing to rotate.

The value that must never ship is `CLIENT_SECRET`, which lives in the backend env and is not in this diff. The constant is annotated in place:

```ts
// Public Discord Application ID (a.k.a. client_id). Safe to ship: it appears in
// every OAuth invite link and is not a secret. Used as the default so the CTA
// works out of the box; override via VITE_DISCORD_CLIENT_ID for a fork.
```

Worth noting the rule is finding this only because the constant was *centralised* by this PR. The same literal was already hardcoded in `Docs.tsx` and inlined in `Landing.tsx` before, so this is not new exposure, it is the same public value in one place instead of two.

Frontend suite green: 86 files, 1019 tests.

PR #1893, opened 2026-07-27
fix(invite): route /invite through the backend and unify the permissions

LucasSantana-Dev:
Follow-up to #1889 / #1888. That PR put the `/invite` redirect in `vercel.json`, which does not apply to production.

## Why the first fix missed

`lucky.lucassantana.tech` does not go through Vercel. It routes through the Cloudflare tunnel to the homelab nginx:

```yaml
ingress:
  - hostname: lucky.lucassantana.tech
    service: http://nginx:80
```

So the canonical invite URL still served the SPA, whose catch-all bounces unknown paths to the landing page. Verified against production after the v2.37.2 deploy:

```
$ curl -o /dev/null -w '%{http_code} %{redirect_url}' https://lucky.lucassantana.tech/invite
200          # no Location header
```

That URL is what the Top.gg listing body, the README CTA and every tracked campaign link in `.agents/tracking-urls.md` point at, so every invite click was swallowed and no `utm_source` ever recorded an install.

## Change

Adds the redirect to `nginx/nginx.conf`, where production actually serves it. `vercel.json` keeps its copy for preview deploys.

- `location = /invite` (exact match) so no other path is caught.
- `302`, not `301`: browsers cache 301 permanently, which would make a wrong `client_id` or permission set unfixable for anyone who had already clicked.
- `permissions=3165184` matches `getBotInviteUrl()` in `lib/discord.ts`. Not Administrator, per the public listing claim.

## Verification

Config validated with `nginx -t` against `nginxinc/nginx-unprivileged:1.31-alpine`, the same base `Dockerfile.nginx` builds from:

```
nginx: the configuration file /etc/nginx/nginx.conf syntax is ok
nginx: configuration file /etc/nginx/nginx.conf test is successful
```

Will re-verify against production once deployed.

## Blocks

The Top.gg resubmission. The listing description links `/invite`, so a reviewer clicking it would land on the homepage rather than an invite prompt.

---
## Summary by cubic
Route `/invite` through the backend to preserve UTM attribution and redirect to Discord OAuth with the minimal permissions. Fixes invite links (Top.gg, README, campaigns) being swallowed by the SPA and stops attribution loss.

- **Bug Fixes**
  - `nginx/nginx.conf`: exact-match `location = /invite` now proxies to the backend so UTM params are logged before redirect; SPA no longer catches it; added notes on why `internal` is not used (Semgrep false positive).
  - Centralized the invite URL/permissions in `@lucky/shared` (`BOT_CLIENT_ID`, `BOT_INVITE_PERMISSIONS=3165184`, `buildBotInviteUrl()`), updated frontend/backend to use it with tests to pin the bits and scope; `vercel.json` now points `/invite` to the backend for previews.

<sup>Written for commit 7c2c9077c3fba1fd48a090f7f067ee4c9997a492. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/1893?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added a dedicated `/invite` link that redirects visitors to the configured Discord authorization page.
  - Ensured consistent invite behavior across frontend and preview deployments.

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -50,6 +50,12 @@ server {
         add_header X-Content-Type-Options nosniff always;
         add_header Referrer-Policy strict-origin-when-cross-origin always;
 
+        # nosemgrep: generic.nginx.security.missing-internal.missing-internal
+        # The rule wants `internal;`, which would make this location
+        # unreachable from outside — this is a public entry point by design,
+        # exactly like the /api and / blocks below. Its SSRF concern does not
+        # apply: the upstream is a fixed literal with no request-derived input
+        # in the URL, so a client cannot steer where the proxy connects.
         set $invite_upstream http://backend:3000;
         proxy_pass $invite_upstream;
         proxy_http_version 1.1;
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -36,6 +36,25 @@ server {
         proxy_read_timeout 300s;
     }
 
+    # Canonical invite URL. It is what the Top.gg listing, the README CTA and
+    # every tracked campaign link point at, so it has to reach Discord rather
+    # than the SPA. The frontend has no /invite route and its catch-all sends
+    # unknown paths to the landing page, which silently swallowed every invite
+    # click (#1888). vercel.json carries the same redirect for preview
+    # deploys, but production is served through the Cloudflare tunnel to this
+    # nginx, so the rule has to exist here too.
+    #
+    # permissions=3165184 is View Channels, Send Messages, Embed Links,
+    # Connect, Speak. NOT Administrator: the public listings state that Lucky
+    # asks for no admin permission. Keep in sync with getBotInviteUrl() in
+    # packages/frontend/src/lib/discord.ts.
+    #
+    # 302, not 301: browsers cache 301 permanently, so a wrong client_id or
+    # permission set would be unfixable for anyone who had already clicked.
+    location = /invite {
+        return 302 "https://discord.com/oauth2/authorize?client_id=962198089161134131&scope=bot%20applications.commands&permissions=3165184";
+    }
+
     location /api {
         set $backend_upstream http://backend:3000;
         proxy_pass $backend_upstream;
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -36,23 +36,27 @@ server {
         proxy_read_timeout 300s;
     }
 
-    # Canonical invite URL. It is what the Top.gg listing, the README CTA and
-    # every tracked campaign link point at, so it has to reach Discord rather
-    # than the SPA. The frontend has no /invite route and its catch-all sends
-    # unknown paths to the landing page, which silently swallowed every invite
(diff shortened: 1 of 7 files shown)


LucasSantana-Dev:
Good catch, and it was more than a P2. Fixed in `834c4537`.

You were right that `backend/src/routes/invite.ts` already logs `[invite] click` with the utm parameters, and that a `return 302` at the edge would have bypassed it. I had checked the frontend router and `vercel.json` and concluded no `/invite` handler existed anywhere — I never grepped the backend. The real bug was narrower than I described: the handler was fine, but nginx only proxies `/api`, so nothing ever reached it.

nginx now proxies `location = /invite` to the backend rather than redirecting itself, so tracked links keep their attribution.

**What that turned up.** Following your pointer to the backend handler surfaced a third permission value:

| source | value | permissions |
|---|---|---|
| `backend/src/routes/invite.ts` | `36970496` | Manage Messages, Connect, Speak |
| `frontend/src/lib/discord.ts` | `3165184` | View Channels, Send Messages, Embed Links, Connect, Speak |
| docs page (fixed in #1889) | `8` | **Administrator** |

The backend value is the worst of the three: it requests Manage Messages (message deletion) while omitting View Channels and Send Messages, so it is simultaneously over-scoped and non-functional. And since it is the handler every tracked link resolves to, it is what real users were being asked to approve — while the Top.gg listing says "No admin permission" and implies a minimal set.

`BOT_INVITE_PERMISSIONS` now lives in `packages/shared/src/constants/invite.ts`, consumed by both the backend redirect and the frontend helper. The test pins the bitfield rather than the string: no Administrator, no Manage Messages, exactly the five bits the bot needs, plus the `applications.commands` scope whose absence caused #1885.

`vercel.json` points at the backend too, so preview deploys log identically.

`nginx -t` clean against `nginxinc/nginx-unprivileged:1.31-alpine`. Suites green: shared 1398, frontend 1019, backend 1350.

LucasSantana-Dev:
False positive, annotated with `nosemgrep` and a comment explaining why (`7c2c9077`).

`internal;` makes a location reachable only from internal redirects. Applying it here would make the invite link return 404 for every real visitor, which is the exact failure this PR exists to fix. It is a public entry point by design — the existing `/api` and `/` blocks are the same shape and neither uses `internal` either.

The rule's SSRF concern requires request-derived input in the proxy target, so that a client can steer where the proxy connects. Here the upstream is a fixed literal:

```nginx
set $invite_upstream http://backend:3000;
proxy_pass $invite_upstream;
```

No part of the request reaches that value. The `set`-then-`proxy_pass` indirection is this repo's existing convention for deferring DNS resolution to the runtime resolver, not a place variables enter the URL.

Issue #1890, opened 2026-07-26
P1: releases are not reaching production — prod stuck on v2.36.1


LucasSantana-Dev:
## Impact

**Production has not received a release since v2.36.1 (2026-07-16).** Both releases since then reported a successful deploy webhook and then failed validation:

| Release | Date | Deploy run | Result |
|---|---|---|---|
| v2.37.0 | 2026-07-22 | `29889191156` | failure |
| v2.37.1 | 2026-07-26 | `30221422334` | failure |

Every fix merged in that window is on `main` and tagged but **not running**.

## Evidence

From run `30221422334`, the `Validate deployed version` step:

```
Attempt 31 ... SHA mismatch (expected=199d91d41a78a68b370b418f776526b5c0c3da2f,
                             actual=597f2cff849de4b31e5d382799e8198ec969fc7e)
...
##[error]Deployed version SHA did not match expected before deadline
```

`597f2cff` is `chore: release 2.36.1 (#1851)`. Production is serving a build two releases old.

## What did and did not happen

Step results for the failing run:

```
3. Resolve deploy SHA:          success
4. Wait for Docker images:      success   <- images for 199d91d4 WERE published
6. Trigger deploy webhook:      success   <- webhook accepted
7. Validate deployed version:   failure   <- never became 199d91d4
```

So this is not a build or registry problem, and the webhook is not rejecting the request.

The containers **did** restart: `GET /api/health` immediately after the run reported `uptime: 206` seconds, matching the deploy window. They just came back up on the old image.

That points at the homelab side pulling or selecting the wrong image rather than at CI — for example `docker compose up -d` without a `pull`, an `IMAGE_TAG` that still resolves to the previous build, or a failed pull silently falling back to the locally cached image. Related prior finding: [docker restart is not the same as recreate](https://github.com/LucasSantana-Dev/Lucky/issues) — `up -d --no-deps` is needed to actually recreate with new env/images.

## Why it matters beyond the version number

The deploy is **failing loudly but late**. The webhook returns success, so the only signal that production did not update is the validation step at the very end. Anyone reading "deploy failed" after a green webhook could reasonably assume a flaky health check and move on — which is likely what happened on 2026-07-22, leaving prod stale for four days without anyone noticing.

## Blocks

- The Top.gg resubmission (#1885). The global slash-command fix shipped in v2.37.1 and is not live, so a reviewer would still find no commands.
- The #1886 guild-command cleanup, which must run only after the new registration code is live.

[...]

PR #1912, opened 2026-07-28
docs(readme): drop test counts and the zero-incident claim

LucasSantana-Dev:
## Test counts

A test count says nothing about whether the bot works. It reads as filler, and it goes stale the moment anyone adds a test — the README said `~6,300` in five places while the suites actually totalled 6,955 (shared 1398, bot 3188, backend 1350, frontend 1019).

Replaced with what a reader is actually deciding on:

| before | after |
|---|---|
| `~6,300 tests · Zero prod incidents` | `No paywall, no premium tier` |
| `Every PR runs lint + build + ~6,300 tests + SonarCloud gates` | `Every PR runs lint, build, the full test suite, and SonarCloud gates` |
| `~6,300 unit + integration tests, Playwright e2e` | `Jest + Vitest unit/integration, Playwright e2e` |
| `# Run all tests (~6,300 total)` | `# Run all tests` |
| `npm run verify (lint + build + ~6,300 tests)` | `npm run verify (lint + build + tests)` |

The suite still runs on every PR. It just does not need a number attached to be worth having.

## The zero-incident claim

`Zero prod incidents` and `Zero incidents: Battle-tested in production with full test coverage` also had to go, because they stopped being true. Production sat on a stale build from 2026-07-16 for four days after two deploys failed their health checks and auto-rolled back without anyone noticing (#1890).

Swapped for the property that is genuinely worth advertising and that a self-hoster actually cares about:

> **Auto-rollback**: A deploy that fails its health checks reverts to the last good build

Verified in `scripts/deploy.sh:465` before claiming it, and it is exactly what kept service up during that incident. It answers "what happens when a deploy of a self-hosted bot goes wrong", which is a better pitch than a coverage boast.

## Note

The same test-count line was in the Top.gg listing (I had just updated it to `~6,900` during the resubmission). Same reasoning applies, so I will strip it there too rather than leave the two out of sync.

---
## Summary by cubic
Remove test counts and all “zero incidents” claims from the README. Replace with durable details: “No paywall, no premium tier”, CI gate (lint, build, full test suite, SonarCloud), test stack (`Jest` + `Vitest`, `Playwright`), “Auto-rollback” on failed deploys, and that incidents are tracked as public GitHub issues with root cause and fix documented.

<sup>Written for commit 5ca1b6c2730dda428df8bedd72901404eb256701. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/1912?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,7 +368,7 @@ Lucky is a **solo personal project** — actively developed and deployed to prod
 - **Automated scanning**: Trivy on every Docker build, Dependabot for dependencies
 - **Merge strategy**: Auto-merge for patch updates, manual review for major changes
 - **Monitoring**: Sentry for error tracking, custom telemetry in production
-- **Incident response**: Zero production incidents to date; post-incident reviews are public
+- **Incident response**: Incidents are tracked as public GitHub issues, with the root cause and fix written up in the open
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 
 <p align="center">
   <b>🎵 The Discord music bot that can't be shut down — because you host it.</b><br>
-  <strong>Self-hosted · Open-source · TypeScript monorepo · ~6,300 tests · Zero prod incidents</strong>
+  <strong>Self-hosted · Open-source · TypeScript monorepo · No paywall, no premium tier</strong>
 </p>
 
 <p align="center">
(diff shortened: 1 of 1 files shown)


LucasSantana-Dev:
Good catch, fixed in `5ca1b6c2`.

I had grepped for `Zero prod incidents` and `Zero incidents` and missed `Zero production incidents to date` in Security & Maintenance — a third phrasing of the same claim, in a section I never opened.

Kept the honest half of that line:

> **Incident response**: Incidents are tracked as public GitHub issues, with the root cause and fix written up in the open

That is a stronger signal than a streak anyway. A streak only holds until it doesn't, and #1890 is the proof — whereas "we write up what broke" stays true and is checkable against the issue tracker.

Verified nothing else remains:

```
$ grep -nEi "zero.{0,15}incident|6,300|6300|~2500" README.md
clean: no test-count or zero-incident claims remain
```

PR #1922, opened 2026-07-29
fix(deploy): make rolled-back deploys visible in failure output

### LINKED ISSUE #1892: ops: a rolled-back deploy looks the same as any other red check
Follow-up from #1890.

## What happened

The v2.37.0 deploy failed on 2026-07-22, auto-rolled back to v2.36.1, and **nobody noticed for four days**. Production silently ran a version behind while `main` kept accumulating merged fixes. It only surfaced because the v2.37.1 deploy failed the same way and got investigated.

## Why it hid

The auto-rollback works correctly, and that is the problem. From the outside:

- the release publishes normally,
- the deploy webhook returns `200`,
- the workflow goes red at the very end on `Validate deployed version`.

A red deploy check reads like a flaky health probe, especially since the auth-config check has a documented history of false-gating. Nothing anywhere says the sentence that matters: **production is not running the version you just released, and has been reverted.**

The rollback message exists, but only inside `/tmp/lucky-deploy.log` on the homelab, inside the webhook container:

```
[deploy] AUTO-ROLLBACK: 199d91d4... failed (health checks failed); reverting to last-good 597f2cff...
[deploy] Deploy failed; auto-rollback restored service on 597f2cff...
```

## Suggested

1. Have the `Notify failure` step distinguish "deploy failed, rollback restored service on `<sha>`" from a generic failure, and say which version production is actually on.
2. Surface the deployed SHA somewhere pollable. The workflow already compares against one; exposing it (for example on `/api/health`) would let a scheduled check alert when the running SHA drifts from the latest tag, independent of whether anyone read a deploy log.
3. Consider failing louder when a release ends with production on an older SHA than the tag that triggered it, since that is a strictly worse state than a deploy that never started.

The value here is not the rollback (that already works). It is that a silent successful rollback should not be indistinguishable from noise.

LucasSantana-Dev:
## Summary

Follow-up to #1890. The v2.37.0 deploy failed, auto-rolled back to v2.36.1, and nobody noticed for four days because a rolled-back deploy read like a generic red check. This PR makes the rollback state explicit:

- **Rollback-aware failure notification**: restores a `Notify failure` step (`if: failure()`) in `deploy.yml`, backed by new `scripts/deploy/notify-deploy-failure.sh`. On any failed deploy it queries `/api/health/version` and annotates the run with the SHA production actually serves: "auto-rollback restored service on `<sha>`" vs. a generic failure, or "failure is in a post-deploy check" when production does run the expected SHA.
- **Loud failure when production ends up older than the release**: the script compares the running SHA against the deploy target via the GitHub compare API; when the running SHA is strictly older (behind_by=0, ahead_by>0) it emits `PRODUCTION IS BEHIND THE RELEASE: <sha> is N commit(s) older than <tag>`. The `Validate deployed version` step's final error now also names the last observed production SHA instead of a bare mismatch.
- **Pollable deployed SHA**: `/api/health` now includes `commitSha` (from the `COMMIT_SHA` env var already baked into the image via the Dockerfile ARG/ENV, same source as `/api/health/version`), so a scheduled check can alert when the running SHA drifts from the latest tag.

Note: the `Notify failure` step referenced in the issue had been removed in #1897; this restores it in rollback-aware form.

## Verification

- `npx jest tests/unit/routes/health.test.ts tests/integration/routes/health.test.ts` in packages/backend: 2 suites, 20 tests passed (includes 2 new `/api/health` commitSha tests).
- `npm run type:check` (backend): clean; full monorepo type:check also ran via pre-commit hook and passed.
- `eslint src/routes/health.ts`: clean (test file is ignored by the repo eslint config, as before).
- `bash -n scripts/deploy/notify-deploy-failure.sh`: OK; deploy.yml parses as valid YAML.
- Script branch behavior exercised locally with stubbed curl/gh: rollback case, post-deploy-check case, and unreachable-health case all emit the intended annotations and exit 0.

Closes #1892

---
## Summary by cubic
Make rolled-back deploys obvious in CI and surface the live commit SHA so silent rollbacks don’t read like generic failures. Failed deploys now say when production auto-rolled back, which SHA is serving, and when the failure happened before rollout.

- **Bug Fixes**
  - Restore a rollback-aware "Notify failure" step in deploy.yml using scripts/deploy/notify-deploy-failure.sh. It reads /api/health/version to annotate the run with the live commitSha, distinguishes auto-rollback vs. post-deploy checks, adds a clear pre-deploy failure message when no SHA is resolved, treats HTTP errors as unreachable, and warns when production is behind the release.
  - Expose commitSha on /api/health and /api/health/version (from COMMIT_SHA) and add tests. The "Validate deployed version" step now prints the expected and last-seen production SHA with a clear hint when they differ.

<sup>Written for commit 07ab6ae20a71250b12c2b55e43c092bbae570382. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/1922?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Health checks now include a deployed commit identifier when available.
  * Deployment failure notifications now distinguish pre-deployment failures from post-deployment verification failures.
* **Bug Fixes**
  * Improved “validate deployed version” timeout reporting by showing the last served commit identifier (when detected) and clarifying rollback vs. flaky checks.
  * Deployment failure diagnostics better differentiate between unreachable health checks and missing commit info.
* **Tests**
  * Added unit test coverage for `/api/health` with `COMMIT_SHA` set and unset.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -294,6 +294,7 @@ jobs:
 
                   deadline=$(($(date +%s) + 600))
                   attempt=0
+                  last_actual=""
 
                   while [ "$(date +%s)" -lt "$deadline" ]; do
                     attempt=$((attempt + 1))
@@ -315,6 +316,7 @@ jobs:
                         echo "==> Deployed SHA matches: $actual"
                         exit 0
                       fi
+                      last_actual="$actual"
                       echo "SHA mismatch (expected=$expected, actual=$actual)."
                     else
                       echo "Not ready (HTTP $http_code)."
@@ -327,7 +329,10 @@ jobs:
                     sleep "$sleep_for"
                   done
 
-                  echo "::error::Deployed version SHA did not match expected before deadline"
+                  echo "::error::Deployed version SHA did not match expected before deadline (expected=$expected, production=${last_actual:-unknown})"
+                  if [ -n "$last_actual" ] && [ "$last_actual" != "$expected" ]; then
+                    echo "::error::Production is running $last_actual instead of $expected. If the homelab auto-rollback restored service, this red check is a silent rollback, not a flaky probe. See the Notify failure step for details."
+                  fi
                   exit 1
 
             - name: Sentry release — finalize
@@ -360,3 +365,17 @@ jobs:
                   EXPECTED_SHA: ${{ steps.resolve_sha.outputs.sha }}
                   REPO: ${{ github.repository }}
               run: bash scripts/deploy/wait-homelab-status.sh wait
+
+            # Runs only when an earlier step failed. Annotates the run with the
+            # SHA production is actually serving so an auto-rollback to the
+            # last-good image is not indistinguishable from a flaky red check
+            # (#1892).
+            - name: Notify failure
+              if: failure()
+              env:
+                  WEBHOOK_URL: ${{ secrets.DEPLOY_WEBHOOK_URL }}
+                  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+                  EXPECTED_SHA: ${{ steps.resolve_sha.outputs.sha }}
+                  RELEASE_TAG: ${{ github.event.release.tag_name }}
+                  REPO: ${{ github.repository }}
+              run: bash scripts/deploy/notify-deploy-failure.sh
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -369,9 +369,10 @@ jobs:
             # Runs only when an earlier step failed. Annotates the run with the
             # SHA production is actually serving so an auto-rollback to the
             # last-good image is not indistinguishable from a flaky red check
-            # (#1892).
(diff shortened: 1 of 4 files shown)


LucasSantana-Dev:
Addressed in 58932449: empty EXPECTED_SHA now reports a plain pre-deploy failure instead of rollback language (plus an `if:` gate separating pre-deploy vs post-rollout notification), `gh api` compare wrapped in `timeout 15`, and reachable-but-no-commitSha is now distinguished from an unreachable endpoint. Deferred as follow-ups: passing an explicit rollout/rollback result back from the homelab (needs homelab-side changes) and the P3 wording on non-main dispatch targets.

Issue #1498, opened 2026-06-18
fix(bot): mod/mgmt commands throw raw Missing Permissions instead of graceful prompt


LucasSantana-Dev:
## Summary
Moderation / management / automod commands throw a **raw Discord `Missing Permissions` API error** to the user when the bot lacks the required permission, instead of a graceful "I'm missing X" prompt. There are **zero** bot-permission checks in the codebase (`grep appPermissions` → 0 hits). With the current minimal-music invite (#1494/#1495 → only View/Send/Embed/Connect/Speak), this is the **default state for every non-music feature on a fresh add**.

## Evidence
- Bot performs actions needing perms it isn't granted on a music-only invite: `ModerateMembers` (timeout), `BanMembers`, `KickMembers`, `ManageRoles`, `ManageChannels`, `ManageGuild`, `ManageMessages`, `ViewAuditLog` (`handlers/auditHandler.ts`).
- `grep -r appPermissions packages/bot/src` → 0 → no command checks the bot's own effective permissions before acting → unhandled `DiscordAPIError[50013] Missing Permissions` surfaces to the user.

## Fix (per ADR `decisions/2026-06-18-invite-permission-scope.md`)
1. Shared guard util: interaction commands check `interaction.appPermissions.has(<needed>)` (discord.js ^14.26) → reply "I'm missing the **X** permission — ask an admin to grant it or re-invite" instead of throwing.
2. Event-driven paths (automod triggers, scheduled mod actions — no interaction): log + skip gracefully, never crash the handler.
3. Pilot on moderation (ban/kick/timeout), then extend to management + automod.

## Acceptance
- A privileged command run while the bot lacks the perm replies with the friendly prompt; no raw API error reaches the user.
- Event-driven privileged action logs + skips when the perm is absent.
- Unit tests for the guard (has-perm → proceeds; missing → prompt/skip).

Prereq for the curated-default invite change in the ADR. Reproduces on `main` (branch-independent).

PR #1927, opened 2026-07-30
fix(invite): adopt the adr curated permission set

### LINKED ISSUE #1923: fix(invite): shipped permission set contradicts the 2026-06-18 invite-scope ADR
Self-reported. I unified the invite permission set in #1893 without checking `decisions/2026-06-18-invite-permission-scope.md`, and landed on a value that ADR explicitly decided against as the end state.

## What the ADR decided

`decisions/2026-06-18-invite-permission-scope.md` (PR #1497, via `/research-and-decide` with a decision-critic pass):

> **Include:** music (View/Send/Embed/Connect/Speak) + `ManageMessages` (cleanup) + `ViewAuditLog` (read-only; needed by `auditHandler`)

It was **sequenced**: ship graceful bot-permission checks first (#1498), *then* adopt the curated default. **#1498 is closed**, so the gate was satisfied and the curated set should be live.

## What is actually shipped

| value | permissions | where |
|---|---|---|
| `3173504` | View, Send, Embed, Connect, Speak, **ManageMessages**, **ViewAuditLog** | what the ADR decided |
| `3165184` | View, Send, Embed, Connect, Speak | **what is live now** (#1893) |
| `36970496` | ManageMessages, Connect, Speak | the old backend value, replaced |

Missing versus the ADR: **`ManageMessages`** and **`ViewAuditLog`**.

## Why this matters beyond tidiness

`auditHandler.ts` reads audit logs to attribute moderation cases. Without `ViewAuditLog` in the invite, every fresh install has that silently degraded until an admin notices and grants it manually. `ManageMessages` is what auto-mod cleanup needs — and #1919 is about the fact that we currently *swallow* the permission error when it is absent, so the two compound: the invite does not request it, and the failure is silent.

The old `36970496` was wrong in a different way (no View/Send at all), so #1893 was still an improvement. It just overshot in the other direction.

## The fix

`packages/shared/src/constants/invite.ts`:

```ts
export const BOT_INVITE_PERMISSIONS = '3173504'
```

Then update `packages/shared/src/constants/invite.spec.ts`, which currently asserts the opposite:

```ts
it('never requests Manage Messages', () => {
    expect(bits & MANAGE_MESSAGES).toBe(0n)   // contradicts the ADR
})
```

That assertion should become the ADR's actual invariant: no `Administrator`, and exactly the seven curated bits.

Also update the public copy, which I changed to match the wrong value:

- Top.gg listing: "Lucky requests only View Channels, Send Messages, Embed Links, Connect and Speak. No Administrator, and no Manage Messages."
- The nginx `/invite` comment in `nginx/nginx.conf`
- `packages/frontend/public/_redirects` if it names the value

"No Administrator" stays true and is the claim worth keeping. "No Manage Messages" has to go.

## Worth confirming first

The ADR is from 2026-06-18. If the intent has since changed — a deliberate decision to keep the invite minimal and let admins escalate — then the right fix is the reverse: amend the ADR to record that, rather than changing the code. **Check with the maintainer before implementing.** That is the one open question here.

LucasSantana-Dev:
Closes #1923. Self-reported: #1893 changed the invite permission set without reading the ADR that governs it.

## What the ADR decided

`decisions/2026-06-18-invite-permission-scope.md` (PR #1497, via `/research-and-decide` with a decision-critic pass):

> **Include:** music (View/Send/Embed/Connect/Speak) + `ManageMessages` (cleanup) + `ViewAuditLog` (read-only; needed by `auditHandler`)

Sequenced behind graceful permission checks (#1498), which is **closed** — so the curated set should already have been live.

## Values, decoded exactly

| integer | permissions | status |
|---|---|---|
| `3173504` | ViewAuditLog, ViewChannel, SendMessages, ManageMessages, EmbedLinks, Connect, Speak | **adopted here** — the ADR set |
| `3165184` | ViewChannel, SendMessages, EmbedLinks, Connect, Speak | shipped in #1893, music-only |
| `36970496` | ManageMessages, UseExternalEmojis, Connect, Speak, UseVAD | old backend value — **no** ViewChannel, **no** SendMessages |
| `8` | Administrator | docs page (#1889) **and `GuildService`** |

Without `ViewAuditLog`, `auditHandler` attribution is silently degraded on every fresh install. That is the concrete cost of the value I shipped.

## The fifth copy nobody had counted

Tracing the call sites turned up `GuildService.generateBotInviteUrl()`:

```ts
const permissions = '8'   // Administrator
```

It backs `GET /api/guilds/:id/invite` (`routes/guilds.ts:87`), which the dashboard calls from `api.ts:182` — the **"add Lucky to this server"** button. So that flow asked every server owner for full Administrator while the Top.gg listing promised the opposite. #1888 and #1894 both counted copies and both missed it.

Now uses the shared constant, like every other entry point.

## Also corrected

`docs/TOP_GG_SUBMISSION.md` specified `36970496` and described it as ten permissions summing to **37022784** — neither that integer nor a set the bot could work with. It also claimed the README used `permissions=8`, removed in #1889. It now points at `/invite` instead of hardcoding an integer, so it cannot drift again.

Dropped its `2500+ tests` and `zero production incidents in 2026` lines too — same claims removed from the README in #1912, and this file is the source for listing copy. Short description re-counted at 108 chars (cap 120).

## Tests

Pinned to the decision rather than to taste:

- no `Administrator`
- exactly the seven curated bits
- none of the six high-alarm permissions the ADR escalates on demand (Kick/Ban/ManageChannels/ManageGuild/ManageRoles/ModerateMembers)

Removed the duplicated literal from `packages/frontend/src/lib/discord.test.ts` — a second place to forget is what caused this. The value is asserted once, in shared.

**Verified the guard bites:** reverting the constant to `3165184` fails the suite (`1 failed, 1403 passed`).

Suites green: shared **1404**, backend **1352**, frontend **1038**. Typecheck clean.

## Follow-up

The live Top.gg listing still says "no Manage Messages". I will correct that copy once this is deployed, so the page and the code agree.

---
## Summary by cubic
Adopted the ADR-curated bot invite permissions across all entry points to stop Administrator prompts and restore audit attribution on fresh installs. Unifies invite URL generation and locks the value in tests and e2e fixtures to prevent drift.

- **Bug Fixes**
  - Set `BOT_INVITE_PERMISSIONS` to `3173504` in `packages/shared/src/constants/invite.ts` (music: View/Send/Embed/Connect/Speak + `ManageMessages` + `ViewAuditLog`) and route all invites through `buildBotInviteUrl()`.
  - Replaced `permissions=8` in `GuildService.generateBotInviteUrl()` with `BOT_INVITE_PERMISSIONS`; the dashboard “add Lucky to this server” no longer asks for Administrator.
  - Updated backend `/invite` to use the shared constant; corrected `docs/TOP_GG_SUBMISSION.md` to link `/invite` and removed outdated integers/claims.
  - Tests and fixtures pin the ADR decision: shared spec asserts the exact bits (no `Administrator`, none of the high‑alarm perms); frontend unit test stops duplicating the literal; e2e fixtures now build URLs from the shared constant (`packages/frontend/tests/e2e/fixtures/test-data.ts`) to avoid masking regressions.

<sup>Written for commit 385ef6ee2db1326e6eb7e8ece7ebe7509d2ff884. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/1927?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Updated bot invitation links to use a curated permission set.
  * Added Manage Messages and View Audit Log permissions where needed.
  * Excluded Administrator and high-risk escalation permissions from initial invitations.
  * Updated submission guidance, descriptions, and setup links.

* **Bug Fixes**
  * Corrected outdated invitation permissions and related documentation.

### DIFF
--- a/docs/TOP_GG_SUBMISSION.md
+++ b/docs/TOP_GG_SUBMISSION.md
@@ -4,23 +4,27 @@ Copy-paste artifacts for filing Lucky on https://top.gg. Launch sequence in `~/.
 
 ## 1. Bot identification
 
-| Field | Value |
-|---|---|
-| Client ID | `962198089161134131` |
-| Invite URL | `https://discord.com/oauth2/authorize?client_id=962198089161134131&scope=bot%20applications.commands&permissions=36970496` |
-| Website | `https://lucky.lucassantana.tech` |
-| GitHub | `https://github.com/LucasSantana-Dev/Lucky` |
-| Support server | *Create a Lucky Discord server first (Phase 0 task #7), then paste invite here.* |
+| Field          | Value                                                                            |
+| -------------- | -------------------------------------------------------------------------------- |
+| Client ID      | `962198089161134131`                                                             |
+| Invite URL     | `https://lucky.lucassantana.tech/invite`                                         |
+| Website        | `https://lucky.lucassantana.tech`                                                |
+| GitHub         | `https://github.com/LucasSantana-Dev/Lucky`                                      |
+| Support server | _Create a Lucky Discord server first (Phase 0 task #7), then paste invite here._ |
 
-**Note on permissions integer `36970496`** — this is Read Messages (1024) + Send Messages (2048) + Manage Messages (8192) + Embed Links (16384) + Attach Files (32768) + Add Reactions (64) + Use External Emojis (262144) + Connect (1048576) + Speak (2097152) + Use Voice Activity (33554432). Scoped — not Admin. Use this for top.gg. The README invite still uses `permissions=8` (Admin) to match the existing Landing page; don't mix the two.
+**Note on the invite URL** — do not hardcode a `permissions=` integer here. `https://lucky.lucassantana.tech/invite` redirects to Discord via the backend, which builds the URL from `BOT_INVITE_PERMISSIONS` in `packages/shared/src/constants/invite.ts`, and logs the `utm_*` parameters on the way through so directory clicks are attributable.
+
+The curated set is `3173504` — View Audit Log, View Channels, Send Messages, Manage Messages, Embed Links, Connect, Speak — per `decisions/2026-06-18-invite-permission-scope.md`. **Never Administrator.** High-alarm permissions (Ban/Kick/ManageRoles/ManageChannels/ManageGuild/ModerateMembers) are escalated on demand rather than requested up front.
+
+_Historical note:_ this file previously specified `36970496` and described it as ten permissions summing to 37022784, which is neither that integer nor a set the bot could work with. The real `36970496` is Manage Messages, Use External Emojis, Connect, Speak, Use Voice Activity — with **no** View Channels and **no** Send Messages, so a bot invited with it could not read or post in a channel. It also claimed the README used `permissions=8` (Administrator); that was removed in #1889.
 
 ## 2. Short description (120 char cap)
 
 ```text
-Self-hosted Discord music bot with autoplay, dashboard, and moderation. TypeScript, open source, 2500+ tests.
+Self-hosted Discord music bot with autoplay, dashboard, and moderation. TypeScript, open source, no paywall.
 ```
 
-Character count: 111.
+Character count: 108.
 
 ## 3. Long description (Markdown supported)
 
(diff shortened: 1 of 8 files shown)


LucasSantana-Dev:
Good catch, fixed in `385ef6ee`.

You are right that it was masking: the fixtures mock the API response, so both invite flows would have kept passing against a hardcoded Administrator URL. They could not have caught the regression this PR fixes, and would not catch a future slip back to `8`.

Both now build from `BOT_INVITE_PERMISSIONS`, with a comment saying why, so they track the constant instead of restating it.

That makes **six** places this value lived. The count kept growing as I traced it — #1888 found three, #1894 a fourth, this PR a fifth in `GuildService.generateBotInviteUrl`, and yours is the sixth. Every one of them is now derived rather than literal, which is the actual fix.

Frontend suite green: 86 files, 1038 tests. Typecheck clean.

PR #1946, opened 2026-08-03
fix(bot): unlink dead Last.fm sessions on error 9 and notify once

LucasSantana-Dev:
## Why

Prod: **2,501 Last.fm "403 Invalid session key (error 9)" lines in 60d** (~42/day), surfaced during the Lavalink re-evaluation. Root cause: the player scrobble path () only warn-logged the failure — the dead  row stayed, so every track produced 2 failing API calls forever. (The external scrobbler already unlinked correctly.)

## What

New shared  () used by **both** scrobble paths, per the critic-reviewed ADR ():

- **DB-row dead key** → unlink + one info log + **one DM** ("relink with /lastfm link"). DM guarded by an in-memory set —  returns true on P2025, so it can't serve as the idempotency signal (double-DM race between updateNowPlaying + scrobble).
- **Env-fallback dead key** (no DB row) → distinct config warning once per process; **never unlink, never DM** (avoids DM-spamming a user who never linked — the critic's critical finding).
- Detection switched from substring  to  — with unlink attached, a WAF HTML 403 would otherwise nuke valid links.
- **Behavior change:** player path now passes  (matching the external scrobbler). Unlinked requesters' plays no longer scrobble to the env  owner's account — that misattribution was latent and activated by unlinking.

## Tests

- New : 5 unit tests (env-key path, DM-once, DM-failure swallow, null client, no-id no-op)
-  updated to the new contract (3 tests rewritten)
- 18/18 touched-suite tests + 99/99 lastfm-suite tests pass; monorepo type-check clean

---
## Summary by cubic
Unlinks invalid Last.fm sessions (error 9) and sends one relink DM, stopping repeated 403s and preventing mis-attributed scrobbles. Uses a shared dead-session handler across the player and external scrobbler.

- **Bug Fixes**
  - Shared `handleDeadLastFmSession`: atomic `unlinkIfKeyMatches(discordId, sessionKey)` with tri-state `'removed' | 'stale' | 'error'`; compare failed key to stored key to protect fresh relinks; log + DM only on `'removed'`.
  - Reliable detection via `isLastFmInvalidSessionError`; `externalScrobbler` now routes invalid-session errors through the shared handler.
  - DM once per session key via bounded LRU; reserve before send and release on failure; delivery is best-effort, at-most-once per key.
  - Env fallback: warn once about `LASTFM_SESSION_KEY`; never unlink or DM; DB lookup failures warn and skip. Player path fallback is requester-scoped (`allowEnvFallback: requesterId === undefined`) to prevent env-account misattribution.

- **Dependencies**
  - Bumped `undici` to `7.29.0`.
  - Audit fixes for `fast-uri` and `ip-address`.

<sup>Written for commit 391db890d69325e18fd401d76490e43f57e542c2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/1946?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of invalid Last.fm sessions during scrobbling and now-playing updates.
  - Automatically removes expired linked sessions and sends at most one relinking message.
  - Prevents repeated notifications and safely handles messaging failures.
  - Protects newer relinked sessions from being removed by stale errors.
  - Avoids fallback account use for identified requesters and logs configuration issues without removing fallback settings.
- **Documentation**
  - Added an architectural decision record documenting invalid-session behavior and trade-offs.


LucasSantana-Dev:
Ran this through /research-and-decide with an adversarial critic pass. Verified directly against `deadSessionHandler.ts`: after a successful unlink the `lastfm_links` row is gone, so any later call for that discordId hits the `if (!row) return` branch and never reaches the notification-guard section again for that session key. The guard-release on a failed send is therefore a defensive no-op under the current call graph, not an active retry path — delivery is best-effort/at-most-once, not guaranteed-once. No code change; clarified the ADR (point 5) and the function's doc comment to state that precisely instead of implying a retry that can't fire. Commit 636a7a6.

PR #1950, opened 2026-08-05
fix(bot): music voice reconnect races — stale-song revert + stacked join/leave

LucasSantana-Dev:
## Summary
- Guard the `connection` event's session-snapshot restore against an active intentional-stop (kick) flag, so a stale pre-kick track can't override a freshly issued `/play`.
- Add a per-guild in-flight recovery lock shared by `checkAndRecover` and the orphan-session monitor, so a flaky connection can't trigger both reconnect paths at once and produce stacked join/leave cycles.

Found via a multi-agent deep-debug session tracing two user-reported symptoms.

Closes #1948
Closes #1949

## Test plan
- [x] `npx jest src/utils/music/watchdog.spec.ts src/handlers/player/lifecycleHandlers.spec.ts` — 29 passed
- [x] Full monorepo `type:check` (ran via pre-commit hook) — clean
- [ ] Manual repro in a test guild: kick bot mid-song, `/play` a different song, confirm it doesn't revert
- [ ] Manual repro: force a flaky voice connection, confirm no stacked join/leave

---
## Summary by cubic
Prevents stale tracks from overriding a new /play after a kick and stops duplicate voice recoveries that caused repeated join/leave loops.

- **Bug Fixes**
  - Guard snapshot restore on connection when an intentional stop (kick) is active, and add a per‑guild recovery lock shared by `checkAndRecover` and the orphan monitor to stop stacked join/leave; tests cover both fixes and verify the lock releases to avoid stuck recoveries.

<sup>Written for commit 3be30e99fe161399f6cbbe87dcd804cd8c35cb38. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/1950?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Issue #1949, opened 2026-08-05
bug(bot): uncoordinated watchdog rejoin + orphan-session monitor cause repeated voice join/leave cycles


LucasSantana-Dev:
## Bug
Bot sometimes joins and leaves a voice channel multiple times in a row during normal playback, with no visible user action triggering it.

## Root cause (confirmed by tracer agent, high confidence)
Two independently-scheduled recovery mechanisms in `packages/bot/src/utils/music/watchdog.ts` can both act on the same guild with no locking/coordination between them, and one of the two performs a double-reconnect on its own:

1. **`checkAndRecover` double-rejoin** (watchdog.ts:157-186): when `queue.connection.state.status !== 'ready'`, it calls `queue.connection.rejoin()`, polls up to `recoveryWaitTimeoutMs` (default 5000ms) for ready; if still not ready, calls `rejoin()` **a second time** and polls again. Each `rejoin()` produces a visible disconnect/reconnect (leave+join) at the Discord voice layer. This fires from the 25s watchdog `arm()` timer (default `MUSIC_WATCHDOG_TIMEOUT_MS`) and also directly from the `disconnect` player event (lifecycleHandlers.ts:127-129).
2. **Orphan session monitor** (`startOrphanSessionMonitor` → `scanOrphanSessions` → `recoverOrphanSession`, watchdog.ts:224-323), on an independent 60s interval, checks every guild with a saved snapshot and — if the existing queue isn't playing — calls `queue.connect(voiceChannel)` (a fresh join, separate from the rejoin path above).

Neither mechanism checks whether the other is already mid-recovery for the same guild. When the underlying voice connection is flaky (network jitter, Discord-side rate limiting, region issues) and stays non-'ready' for more than the ~5-10s covered by `checkAndRecover`'s own double-rejoin, the 60s orphan monitor can fire on top of it, stacking a 3rd/4th join/leave cycle.

Additional finding: `startPeriodicScan`/`scanOrphanedSessions` (watchdog.ts:333-378) is defined but has zero call sites outside its own spec file — dead code, not part of the active bug, but worth removing or wiring up for clarity in the eventual fix.

## Fix direction (not yet implemented)
[...]

Issue #1929, opened 2026-07-30
observability: bot reports healthy while music is completely broken


LucasSantana-Dev:
Surfaced by a `/play` failure on 2026-07-30. The immediate cause was a homelab network outage, not code — but the bot gave no signal at all, which is the part worth fixing.

## What happened

The bot booted at 12:21 while the homelab could not resolve `www.youtube.com`:

```
[YOUTUBEJS][Session]: Failed to retrieve session data from server.
                      Session data generated locally will be used instead.
[YOUTUBEJS][Session]: Failed to retrieve config data. [TypeError: fetch failed]
  [cause]: Error: getaddrinfo EAI_AGAIN www.youtube.com
[12:21:27] YoutubeExtractor registration returned null (attempt 1/2)
[12:21:33] Registered YoutubeExtractor (SoundCloud bridge + YouTube fallback)
```

The #1468 retry did its job — attempt 2 returned a truthy handle, so we logged **"Registered YoutubeExtractor"**. But `youtubei.js` had already fallen back to locally-generated session data because it could not reach YouTube. The extractor exists and is non-functional.

From there:

- `docker inspect lucky-bot` → **`healthy`**
- `/healthz` → **200**, because it only checks `client.isReady()` (the Discord gateway), which was fine
- `/play` → fails for the user
- Nothing in the logs after startup says anything is wrong

So every signal we have said the bot was fine while its primary feature was dead.

## The gap

`Registered YoutubeExtractor` is logged on a truthy return value, which does not mean the extractor can actually resolve anything. We treat "registered" as "working".

## What NOT to do

**Do not add a YouTube reachability probe to the container healthcheck.** That would restart-loop the bot during any YouTube-side outage, turning a degraded-music incident into a full bot outage. The healthcheck is correct to track only the gateway.

## Options worth weighing

1. **Post-registration smoke resolve.** After registering, attempt one cheap known-good search. On failure, log at error level and set a flag. Cost: one extra request at boot; needs care not to slow startup.
2. **Surface it to the user at command time.** When `/play` finds no results *and* the extractor is in a known-degraded state, say "music sources are currently unreachable" rather than "No results found". A wrong error message sent users to #1468 last time.
3. **Expose it as a metric.** `lucky_music_extractor_degraded{extractor="youtube"}` scraped by the homelab Prometheus, alerting after N minutes. Fits the alerting work in #1917 and does not risk restart loops.

[...]

PR #1958, opened 2026-08-10
fix(bot): stop autoplay overriding clear/kick, race in watchdog recovery

### LINKED ISSUE #1948: bug(bot): play after kick reverts to old song after ~5s (session restore ignores intentional-stop flag)
## Bug
Bot playing song X. User kicks bot mid-song. User runs `/play` with different song Y. Bot joins, starts Y, then ~5s later reverts to X.

## Root cause (confirmed by tracer agent, high confidence)
`packages/bot/src/handlers/player/lifecycleHandlers.ts:34-92`, the `connection` event handler, calls `musicSessionSnapshotService.restoreSnapshot(queue, ...)` (line ~69-75) **unconditionally** whenever a voice connection is (re)established — it never checks `musicWatchdogService.isIntentionalStop(queue.guild.id)`.

Contrast with the `disconnect` handler in the same file (line 120-130), which correctly gates `checkAndRecover` behind `!musicWatchdogService.isIntentionalStop(guildId)`.

## Sequence
1. Kick → `voiceStateUpdate` (lifecycleHandlers.ts:11-23) calls `markIntentionalStop(guildId)`, flag held 35s (`timeoutMs+10_000`).
2. `disconnect` event saves a snapshot with song X as `currentTrack` (line 126), correctly skips `checkAndRecover` (line 127-129, guarded).
3. `/play` reconnects → `connection` event fires (line 34) → restores the stale snapshot (song X) **without checking the intentional-stop flag**, per `sessionSnapshots.ts` restore logic which enqueues + plays the restored track when the queue is otherwise empty.
4. Song Y (from `/play`) races against the restored X — X wins playback shortly after Y starts, producing the observed revert.

## Fix direction (not yet implemented)
Guard the `restoreSnapshot` call in the `connection` handler (lifecycleHandlers.ts ~line 67) with the same `!musicWatchdogService.isIntentionalStop(queue.guild.id)` check used in the `disconnect` handler — a kick/leave should not silently reintroduce the old queue on next reconnect.

## Repro
- Have bot playing music in a voice channel.
- Kick the bot from the channel (server mod action, not `/leave`).
- Immediately `/play` a different song.
- Observe: new song plays, then reverts to the pre-kick song ~5s later.

Filed per multi-agent deep-debug session, tracer agent report attached in investigation (not pasted here — see this issue's linked PR/commit when fixed for full trace).

### LINKED ISSUE #1957: bug(bot): autoplay silently undoes /clear and bypasses suppression guards on emptyQueue
## Bug
With autoplay on, running `/clear` (and, separately, the natural queue-drain path) gets silently undone by autoplay refilling the queue — the user's `/clear` has no lasting effect while a track keeps playing, matching the "autoplay overrides clear queue" report.

## Root cause (confirmed by direct code read, high confidence)

Two separate gaps, both in the music-playback path:

1. **`packages/bot/src/functions/music/commands/clear.ts`** never calls into the autoplay-suppression system at all. Contrast with `stop.ts:31` and `leave.ts:41`, which both call `musicWatchdogService.markIntentionalStop(guildId)` before clearing/leaving. `/clear` empties `queue.tracks` but leaves `repeatMode` and the watchdog/suppression state untouched, so when the currently-playing track ends naturally, nothing tells autoplay the user just asked for an empty queue.

2. **`packages/bot/src/handlers/player/lifecycleHandlers.ts`**'s `emptyQueue` player-event handler (around line 111, pre-fix) calls `replenishQueue(queue)` directly whenever `repeatMode === 3 && queue.currentTrack`, with **no check** of `musicWatchdogService.isIntentionalStop()` or `isReplenishSuppressed()`. This is the same condition the guarded path checks: `playerFinish`/`playerSkip` → `handleQueueExhaustion()` (`packages/bot/src/handlers/player/queueExhaustion.ts:20`) → `replenishIfAutoplay()` (`packages/bot/src/handlers/player/trackHandlers.ts:259-266`), which DOES check both guards via `isReplenishSuppressed`. Two mechanisms do the same job — refill on empty queue — and only one respects suppression. Same uncoordinated-duplicate-mechanism shape as #1949.

Net effect: even if `/clear` had set a suppression flag, the `emptyQueue` path would ignore it and refill anyway.

## Fix (implemented in the accompanying PR)
- `clear.ts`: call `setReplenishSuppressed(guildId, 35_000)` after `queue.clear()` — same window `queueExhaustion.ts` already uses when the queue runs fully dry, reusing the existing suppression store instead of the heavier `markIntentionalStop` (which also disables watchdog reconnect for 35s — too broad for a simple `/clear`).
- `lifecycleHandlers.ts`: gate the `emptyQueue` handler's `replenishQueue()` call behind the same `isIntentionalStop` / `isReplenishSuppressed` checks the guarded path uses.

## Repro
- Enable autoplay (`repeatMode === 3`), have a track playing with a queue behind it.
- `/clear`.
- Let the current track finish naturally.
- Observe: autoplay immediately refills the queue, as if `/clear` never happened.

Filed during a multi-agent debug session investigating a 6-month-old "autoplay overrides commands" report (also produced #1948, #1949 as related-but-distinct root causes on the same complaint).

LucasSantana-Dev:
## Summary

Root-caused the "autoplay overrides commands / bot comes back on its own after disconnect" reports (open ~6 months).

**Reconciled with #1950** (open since 2026-08-05, not yet merged): it already fixes #1948 and #1949 with an equivalent implementation (same guard, same per-guild lock pattern, independently converged). Rather than ship a competing duplicate, this PR now adopts #1950's `lifecycleHandlers.ts`/`watchdog.ts` verbatim and adds only what #1950 doesn't touch:

- `/clear` never suppressed autoplay replenish (unlike `/stop`/`/leave`) — the queue silently refilled once the current track ended. `clear.ts` now calls `setReplenishSuppressed()`, reusing the window `queueExhaustion.ts` already uses.
- The `emptyQueue` player-event handler called `replenishQueue()` directly, bypassing the `isIntentionalStop`/`isReplenishSuppressed` guards its sibling `playerFinish`/`playerSkip` path respects. Now checks both — filed as #1957.

**Once #1950 merges**, this PR's diff on `lifecycleHandlers.ts`/`watchdog.ts` should be empty (rebase will show no conflict) and only the `clear.ts` + `emptyQueue` fix for #1957 remains as net-new content. Recommend merging #1950 first.

YouTube extractor: code itself checked clean (search/playlist/error-handling paths). The "extractor has issues" complaint maps to already-open #1929 (extractor silently degrades — registers truthy but non-functional on YouTube-side session failures); its fix (Prometheus metric + user-facing wording) is a distinct, larger scoped task, left as a follow-up rather than bolted on here.

## Blocked on a repo-wide CI issue (not this PR)

Both this PR and #1950 currently fail the required **Security** check — this is **not caused by either PR's diff**. Root cause: `package.json`'s `overrides` pins `undici` to exactly `7.28.0` (itself now a vulnerable version — advisory range `7.0.0–7.28.0`) and floors `ip-address` at `>=10.1.1` with no ceiling past its vulnerable range (`<=10.3.0`). Confirmed pre-existing on clean `main` via `git stash` + `node scripts/audit-gate.mjs`.

I attempted a fix (bump the override pins to `undici: 7.29.0`, `ip-address: >=10.3.1`, add `fast-uri: >=3.1.5`) but `npm install` did not consistently re-resolve the root-level packages to the new floors even after a full reinstall, and the local tree showed additional drift (`nanoid` findings appearing/disappearing between runs, a pre-existing `react-router` advisory-id mismatch unrelated to my changes). This needs a careful, dedicated lockfile-health pass with full CI verification — not something to force through inline here. Did not commit a half-verified `package-lock.json`.

Closes #1957

## Test plan
- [x] `npx jest clear.spec.ts lifecycleHandlers.spec.ts watchdog.spec.ts` — 50 passed (watchdog/lifecycleHandlers tests are now #1950's, adopted as-is; new emptyQueue suppression tests added)
- [x] `tsc --noEmit` clean on all touched files
- [x] `eslint` clean

### DIFF
--- a/packages/bot/src/functions/music/commands/clear.spec.ts
+++ b/packages/bot/src/functions/music/commands/clear.spec.ts
@@ -15,6 +15,7 @@ const createErrorEmbedMock = jest.fn((title: string, desc?: string) => ({
 const resolveGuildQueueMock = jest.fn()
 const debugLogMock = jest.fn()
 const errorLogMock = jest.fn()
+const setReplenishSuppressedMock = jest.fn()
 
 jest.mock('../../../utils/command/commandValidations', () => ({
     requireGuild: (...args: unknown[]) => requireGuildMock(...args),
@@ -32,6 +33,11 @@ jest.mock('../../../utils/general/embeds', () => ({
 
 jest.mock('../../../utils/music/queueResolver', () => ({
     resolveGuildQueue: (...args: unknown[]) => resolveGuildQueueMock(...args),
+}))
+
+jest.mock('../../../utils/music/replenishSuppressionStore', () => ({
+    setReplenishSuppressed: (...args: unknown[]) =>
+        setReplenishSuppressedMock(...args),
 }))
 
 jest.mock('@lucky/shared/utils', () => ({
@@ -115,6 +121,33 @@ describe('clear command', () => {
         } as any)
 
         expect(queue.clear).toHaveBeenCalled()
+    })
+
+    it('suppresses autoplay replenish after clearing so it cannot silently refill the queue', async () => {
+        const queue = createQueue(5)
+        resolveGuildQueueMock.mockReturnValue({ queue })
+
+        await clearCommand.execute({
+            client: {} as any,
+            interaction: makeInteraction(),
+        } as any)
+
+        expect(setReplenishSuppressedMock).toHaveBeenCalledWith(
+            'guild-1',
+            expect.any(Number),
+        )
+    })
+
+    it('does not suppress replenish when queue was already empty', async () => {
+        const queue = createQueue(0)
+        resolveGuildQueueMock.mockReturnValue({ queue })
+
+        await clearCommand.execute({
+            client: {} as any,
+            interaction: makeInteraction(),
+        } as any)
+
+        expect(setReplenishSuppressedMock).not.toHaveBeenCalled()
     })
 
     it('responds with success message and count', async () => {
--- a/packages/bot/src/functions/music/commands/clear.ts
+++ b/packages/bot/src/functions/music/commands/clear.ts
@@ -1,5 +1,8 @@
 import { SlashCommandBuilder } from '@discordjs/builders'
-import { createErrorEmbed, createSuccessEmbed } from '../../../utils/general/embeds'
+import {
+    createErrorEmbed,
+    createSuccessEmbed,
+} from '../../../utils/general/embeds'
 import { interactionReply } from '../../../utils/general/interactionReply'
 import { debugLog, errorLog } from '@lucky/shared/utils'
 import Command from '../../../models/Command'
@@ -11,6 +14,12 @@ import type { CommandExecuteParams } from '../../../types/CommandData'
 import type { ChatInputCommandInteraction } from 'discord.js'
 import type { GuildQueue } from 'discord-player'
 import { resolveGuildQueue } from '../../../utils/music/queueResolver'
+import { setReplenishSuppressed } from '../../../utils/music/replenishSuppressionStore'
+
+// Same window queueExhaustion.ts uses when the queue runs dry with nothing
+// left to recover — keeps /clear from being silently undone by autoplay
+// refilling the queue the moment the current track ends.
+const CLEAR_REPLENISH_SUPPRESSION_MS = 35_000
 
 async function handleEmptyQueue(
     interaction: ChatInputCommandInteraction,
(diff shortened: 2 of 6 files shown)


LucasSantana-Dev:
Superseded by #1998 (merged) — same recovery-lock + suppression-flag guards already landed there (this branch's own reconcile commit already noted the overlap with #1950). Closing as redundant.

Issue #2002, opened 2026-08-12
ci: docker-build GHA cache scope is service-only, not branch-scoped — corrupts under concurrent PR builds


LucasSantana-Dev:
## Context

Surfaced while merging a queue of ~14 open PRs, several with Docker CI running concurrently.

## Problem

`.github/actions/docker-build-service/action.yml` (used by `.github/workflows/ci.yml`'s `docker-build` matrix job) sets:

```yaml
cache-from: type=gha,scope=${{ inputs.service }}
cache-to: type=gha,mode=max,scope=${{ inputs.service }}
```

The scope is just the service name (`bot`, `backend`, `frontend`, `nginx`) — shared across every PR branch and `main`, not branch/ref-scoped. `docker-publish.yml` (the main-push image build) has the same pattern.

When two or more PRs' Docker builds run concurrently (e.g. during a merge-train / bot-PR-queue drain), they write `mode=max` cache exports to the same GHA cache scope at the same time. This produces a corrupted/dangling cache manifest — observed failure:

```
ERROR: failed to build: failed to solve: failed to compute cache key: failed to calculate checksum of ref <hash>::<hash>: "/app/packages/backend/node_modules": not found
```

Reproduced twice in this session: cleared all `index-bot-*`/`index-backend-*`/`buildkit-blob-*` GHA caches (478 stale entries existed — also worth noting as its own cleanup signal), then re-ran two PRs' Docker builds within seconds of each other — the corruption recurred immediately with a fresh cache-ref hash.

## Suggested fix

Scope the cache key to include something that prevents cross-branch/cross-concurrent-run collisions, e.g.:

```yaml
cache-from: |
    type=gha,scope=${{ inputs.service }}-${{ github.ref_name }}
    type=gha,scope=${{ inputs.service }}-main
cache-to: type=gha,mode=max,scope=${{ inputs.service }}-${{ github.ref_name }}
```

(import from both the branch's own scope and `main`'s as a fallback, export only to the branch's own scope — avoids two branches writing the same key while still getting `main`'s warm cache on first run). Alternatively, add a repo-level `concurrency:` group across all `docker-build-*` jobs regardless of PR, serializing Docker builds entirely (simpler, costs wall-clock time on a busy queue).

## Repro

Not a one-off flake — reproduced by running 2+ PRs' `docker-build` matrix jobs concurrently. Single-PR builds (no concurrent Docker CI elsewhere) do not appear to hit this.

[...]

PR #2006, opened 2026-08-12
fix(ci): extract node_modules before per-workspace build steps run

LucasSantana-Dev:
## Summary

Follow-up to #2005. That PR split \`deps-production\` per production target, which fixed \`Build — bot\` (it no longer needs the backend build step to run at all, since nothing bot's target consumes depends on it) but \`Build — backend\` kept failing identically — confirming the bug tracks the backend build step's execution, not which node_modules directory gets copied.

## Root cause

\`COPY --from=<stage>\` always waits for that stage's LAST instruction to complete, even when the specific path being copied was already final much earlier. \`node_modules\` is fully installed by \`npm ci\`, long before any per-workspace build script runs — none of those scripts write into \`node_modules\`. But because \`packages/backend/node_modules\` is copied from the \`build\` stage (whose last instruction is \`RUN npm run build --workspace=packages/backend\`), every COPY of it implicitly depends on that RUN completing — and something about that specific RUN step intermittently corrupts BuildKit's cache-key resolution for the immediately-following COPY.

## Change

Adds an \`installed-deps\` checkpoint stage right after \`npm ci\`, before any per-workspace \`RUN\` build step. \`deps-production-base\`/\`-bot\`/\`-backend\` now copy \`node_modules\` from this checkpoint instead of the final \`build\` stage — fully decoupling the node_modules COPY from whatever the backend build RUN step does. Behavior-neutral: same files, same content, just sourced from an earlier point in the same build graph.

## Test plan

- [ ] This PR's own \`Build — bot\`/\`Build — backend\`/\`Build — Docker images\` checks pass
- [ ] Once merged, #1956 and #1952 sync and build cleanly (both bot and backend)

---
## Summary by cubic
Decouples Docker COPY sources from later RUNs to avoid intermittent BuildKit checksum failures. Previously production images copied from a monolithic `build` stage tied to per-workspace builds; now each COPY pulls from the earliest stage that produces the path. Behavior-neutral; still compiles `@discordjs/opus` once.

- Add an empty `installed-deps` checkpoint right after `npm ci`; `deps-production-*` now copy each workspace’s `node_modules` from it.
- Introduce `source-copied` (source + `prisma generate`) and split builds into `build-shared` → `build-bot`, `build-backend`, `build-frontend`; copy shared/prisma/dist from these stages instead of `build`.
- Make `build-frontend` inherit from `build-shared` so `@lucky/shared/constants` resolves; copy `@prisma` engines and `prisma` from `source-copied`.
- Copy `packages/shared/src/generated` from `source-copied` (where `prisma generate` writes it) instead of `build-shared`.

<sup>Written for commit bed52ee68aa8bc93018d8f76b41e7723cd604025. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/2006?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Refactor**
  * Improved the application build process with more efficient, parallelizable stages.
  * Streamlined production packaging for frontend, backend, bot, and database components.
  * Preserved existing dependency-pruning behavior.

### DIFF
--- a/.github/actions/docker-build-service/action.yml
+++ b/.github/actions/docker-build-service/action.yml
@@ -17,32 +17,14 @@ inputs:
         default: 'false'
     use-cache:
         description: >-
-            Import AND export the gha layer cache for this service. Set to
-            'false' on retries: BuildKit's "failed to calculate checksum of
-            ref ...: not found" failure persists even with cache-from alone
-            disabled (confirmed via repeated fresh-hash reruns), which points
-            at cache-to instead — `mode=max` exports every intermediate
-            layer as the build progresses, and a later stage's COPY --from
-            can race against that same layer still being finalized for
-            export. Disabling cache-to too on retries removes that export
-            pressure entirely for a plain, uncached build.
+            Import the gha layer cache for this service. Set to 'false' on
+            retries: BuildKit's "failed to calculate checksum of ref ...:
+            not found" failure comes from a stale/corrupted gha cache entry,
+            and cache-from is scoped identically across retries within the
+            same job, so a retry that reimports it fails identically instead
+            of getting a fresh build.
         required: false
         default: 'true'
-    driver:
-        description: >-
-            Buildx driver. Retries pass 'docker' (the classic engine's
-            built-in BuildKit, no separate container) instead of the default
-            'docker-container'. The "failed to calculate checksum of ref
-            ...: not found" flake is BuildKit's own snapshot/cache-key
-            resolution failing inside that separate container's snapshotter
-            (confirmed present across GC-pressure, gha cache-from, gha
-            cache-to, and a pinned older BuildKit daemon build — see PR
-            history on this file). 'docker' uses a different snapshotter
-            entirely, sidestepping the buggy code path. Loses layer caching,
-            which retries already forgo via use-cache=false, so nothing
-            extra is given up.
-        required: false
-        default: 'docker-container'
 runs:
     using: composite
     steps:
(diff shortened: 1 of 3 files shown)


LucasSantana-Dev:
## The node_modules decoupling didn't address the failures actually being hit

Since this PR's checkpoint-stage fix landed, I've been testing further mitigations on this branch (BuildKit daemon version pin, buildx driver fallback — both disproven, see #2002). Looking closer at the actual failure logs from every attempt on this branch (not just this PR's premise): the failing \`COPY --from\` paths are consistently

\`\`\`
"/app/packages/shared/src/generated": not found
"/app/packages/shared/dist": not found
"/app/packages/backend/dist": not found
"/app/prisma": not found
\`\`\`

None of these is \`node_modules\`. They're build *outputs* — copied from the \`build\` stage's \`deps-production-*\` COPY instructions, which necessarily run after the per-workspace \`RUN npm run build\` steps (you can't decouple a build output from the RUN that produces it the way \`installed-deps\` decouples node_modules, which exists before any RUN).

This PR's root-cause theory ("RUN npm run build --workspace=packages/backend intermittently corrupts BuildKit's cache-key resolution for the immediately-following COPY") predicted the node_modules COPY was the affected one — that's now decoupled and no longer implicated in any of this session's ~10 build attempts on this branch. But the dist/generated/prisma COPYs, which still directly follow build RUN steps in the \`build\` stage, are exactly what's failing every time.

This matches the upstream bug's shape (moby/buildkit#4128: COPY --from referencing a stage whose last instruction was a RUN, resolved after that RUN completes) more precisely than the original theory did — it's not specific to the backend build RUN corrupting a *specific* later COPY, it's that *any* COPY --from immediately following *any* RUN in that stage is at risk, and this Dockerfile still has several (shared build → shared dist/generated COPY, backend build → backend dist COPY, plus the untouched prisma generate step → prisma COPY).

Not attempting a further Dockerfile restructure on this myself right now — decoupling *these* COPYs would need a checkpoint-per-build-output pattern (one intermediate stage after each RUN, mirroring what installed-deps did for node_modules), which is a real structural change to this Dockerfile and deserves a proper look with fresh eyes rather than another speculative patch stacked onto an already-long investigation. Full history on #2002.

LucasSantana-Dev:
Fixed per cubic's review: shared/src/generated now sources from `source-copied` (right after `npx prisma generate`, which is what actually writes it — `prisma/schema.prisma` output is `../packages/shared/src/generated/prisma`) instead of `build-shared`.

Issue #2051, opened 2026-08-21
fix(bot): Spotify extractor registration failure is invisible — no degraded gauge, no Sentry


LucasSantana-Dev:
## Summary

`registerSpotifyExtractor()` catches its failures and emits only a `warnLog`. Unlike the YouTube extractor it never calls `setExtractorDegraded('spotify', ...)`, so there is no Prometheus gauge, no in-process health state, and nothing readable at error time. A bot running without a working Spotify extractor is indistinguishable from one where "the artist genuinely has no tracks".

## Evidence

`packages/bot/src/handlers/player/playerFactory.ts:70-91`

```ts
const registerSpotifyExtractor = async (player: Player): Promise<void> => {
    try {
        const registered = await player.extractors.register(SpotifyExtractor, { ... })
        if (!registered) {
            warnLog({ message: 'SpotifyExtractor registration returned null — Spotify searches degraded' })
            return                       // <-- no setExtractorDegraded
        }
        infoLog({ message: 'Registered SpotifyExtractor (priority 1)' })
    } catch (error) {
        warnLog({ message: 'SpotifyExtractor failed to register — Spotify searches degraded', error })
    }                                    // <-- no setExtractorDegraded
}
```

Compare `loadYoutubeExtractor()` at `:141-201`, which calls `setExtractorDegraded('youtube', true|false)` on every path and escalates to `errorLog` after retries.

`packages/bot/src/handlers/player/extractorHealth.ts` already provides the single-source-of-truth helper and the `musicExtractorDegradedGauge`; Spotify simply is not wired into it.

## Proposed fix

1. Call `setExtractorDegraded('spotify', true)` on both failure paths and `setExtractorDegraded('spotify', false)` on success.
2. Escalate the terminal failure from `warnLog` to `errorLog` so it reaches Sentry, matching the YouTube precedent and #2047 (dead Last.fm env session key).
3. Optionally apply the same retry loop the YouTube path uses.

## Notes

Pre-existing on clean `main` — not a regression from any current branch. Surfaced while diagnosing `/artist` returning "No tracks found for artist queen".

LucasSantana-Dev:
## Production evidence (7d Loki, `container_name="lucky-bot"`)

Quantified against the homelab Loki (`ssh homelab`, `127.0.0.1:3100`). Labels discovered via `/loki/api/v1/labels`, not guessed.

**Registration is healthy — this issue is purely about the missing signal, not a live failure:**

[...]

Issue #2050, opened 2026-08-21
fix(bot): /album silently reports "No tracks found" when the Spotify search arm fails


LucasSantana-Dev:
## Summary

`/album` hard-pins `QueryType.SPOTIFY_SEARCH` for every non-URL query with **no fallback arm**. When the Spotify search returns nothing, the user gets `No tracks found for **<query>**` regardless of how well-known the album is.

Same root cause as the `/artist` report (fixed on `fix/soundcloud-fallback-duration-match`); `/album` was left untouched because its result handling additionally requires `searchResult.playlist`.

## Evidence

`packages/bot/src/functions/music/commands/album.ts:113-138`

```ts
const isUrl = isSpotifyAlbumUrl(query)
const searchEngine = isUrl ? QueryType.AUTO : QueryType.SPOTIFY_SEARCH
const searchResult = await client.player.search(searchQuery, {
    requestedBy: interaction.user,
    searchEngine,
})

if (!searchResult?.tracks.length) {
    // "No tracks found for **<query>**"
}
```

Why the Spotify arm can silently yield zero tracks — `node_modules/discord-player-spotify/dist/index.js:280-325`:

```js
async search(query) {
  try { /* credentialed REST or anon GraphQL */ }
  catch { return null }      // <-- every failure collapses to null
}
```

and `handle()` at `:823`:

```js
const data = await this.internal.search(query);
if (!data) return this.createResponse();   // empty tracks[]
```

So a dead `SPOTIFY_CLIENT_ID`/`SPOTIFY_CLIENT_SECRET`, a 429, or a rotated anon-token flow all present to the user as "this album does not exist", and nothing is logged or sent to Sentry.

Verified locally (2026-08-21): the anon GraphQL `searchDesktop` path is alive and returns Queen tracks, so the library itself is not structurally broken — the arm is environment-dependent, which is exactly why it needs a fallback.

## Proposed fix

[...]

PR #2052, opened 2026-08-21
fix(bot): fall back past a dead spotify arm in /artist

LucasSantana-Dev:
## Problem

`/artist queen` replied `No tracks found for artist **queen**`.

`/artist` hard-pinned `QueryType.SPOTIFY_SEARCH` with **no fallback arm** (`artist.ts:112`, pre-fix). One provider, one attempt. Any Spotify hiccup became "this artist does not exist".

Underneath, `discord-player-spotify` collapses every failure to an empty result with no log:

```js
async search(query) { try { /* ... */ } catch { return null } }   // dist/index.js:280-325
const data = await this.internal.search(query)
if (!data) return this.createResponse()                          // empty tracks[]
```

`/play` already learned this lesson (`3e76068`, "default text search now uses AUTO_SEARCH instead of SPOTIFY_SEARCH ... without requiring Spotify credentials to succeed first"). `/artist` and `/album` never got the same treatment.

## Production evidence

7d Loki (`container_name="lucky-bot"`):

| Signal | 7d |
|---|---|
| `Registered SpotifyExtractor` | 11 |
| `SpotifyExtractor failed to register` / `returned null` | 0 |
| `Failed to retrieve access token` / `Failed to fetch Spotify data` | 0 |
| `Primary search failed, falling back to YouTube` (search layer) | 4 |

Spotify auth and registration are healthy, so this is **not** a credential outage. It is a transient-zero-result that `/artist` had no way to absorb. `/play` absorbs the identical event — `2026-08-21T16:20:09Z`:

```
Primary search failed, falling back to YouTube
{ "query": "Inteligência Marginal 3", "searchEngine": "spotifySearch",
  "error": "NoResultError: No results found ... (Extractor: com.discord-player.itsmaat.spotifyextractor)" }
→ Bridge: streamed via yt-dlp → Started playing "INTELIGÊNCIA MARGINAL #3..."
```

Same class of blip. `/play` recovered and played the track; `/artist` would have told the user the artist does not exist.

The reported `/artist` invocation produced **zero** log lines in 7 days, which is why the exact trigger is unrecoverable.

## Change

- `/artist` iterates `SPOTIFY_SEARCH → YOUTUBE_SEARCH → SOUNDCLOUD_SEARCH`, reusing the `TEXT_SEARCH_BLOCKED_EXTRACTORS` list `/play` already relies on so Spotify/Attachment extractors cannot intercept the text-search arms (#1930).
- Every empty or throwing arm `warnLog`s, so a recurrence is queryable: `{container_name="lucky-bot"} |= "Artist search arm"`. Previously nothing was emitted at all.
- `No results` is reported only after all three arms are exhausted.
- Playback follows the arm that resolved: `SPOTIFY_SONG` for Spotify, `AUTO` otherwise, so a YouTube/SoundCloud URL is not re-searched against Spotify.
- `TEXT_SEARCH_BLOCKED_EXTRACTORS` exported from `play/handlers/resolveProvider.ts` (no new abstraction, no behavior change for `/play`).

## Tests

4 new tests in `artist.spec.ts`. Verified all 4 fail on the pre-fix code and pass after:

1. falls back past an empty Spotify arm instead of reporting "no tracks found"
2. plays a non-Spotify fallback track with `AUTO` rather than `SPOTIFY_SONG`
3. reports no results only after every arm is exhausted
4. continues to the next arm when an arm rejects

Monorepo typecheck clean; 144 tests pass across the artist + play suites.

## Follow-ups (filed, not in this PR)

- #2050 — `/album` has the identical defect. Left out because its `!searchResult.playlist` branch needs a design call: text search on YouTube/SoundCloud returns no playlist.
- #2051 — Spotify extractor failures never reach Sentry. `LogService.warn()` only calls `addBreadcrumb`; solely `error()` calls `captureMessage` (`log/service.ts:140-167`), and registration failures use `warnLog`.

---
## Summary by cubic
Makes /artist resilient to Spotify search failures and keeps artist-only queues. Previously it pinned SPOTIFY_SEARCH and dropped the artist filter unless ≥3 matches; now it falls back to YouTube/SoundCloud, trusts exact author matches at any count, and reports “No results” only after all arms fail.

- Search fallback: iterates SPOTIFY_SEARCH → YOUTUBE_SEARCH → SOUNDCLOUD_SEARCH; reuses `TEXT_SEARCH_BLOCKED_EXTRACTORS` to prevent extractor interception; warn-logs once per arm (thrown vs empty); “No results” only after all arms fail.
- Playback: uses SPOTIFY_SONG when the Spotify arm resolves; uses AUTO for non-Spotify results to avoid re-searching resolved URLs.
- Artist filter: exact author matches kept regardless of count; keep ≥3 for word/substring matches; fall back to provider order only when zero tracks match the artist.
- Shared: exports `TEXT_SEARCH_BLOCKED_EXTRACTORS` from `packages/bot/src/functions/music/commands/play/handlers/resolveProvider.ts`; `/play` unchanged.
- Tests: 9 cases covering fallback across arms, single-log semantics, Spotify vs non-Spotify playback, correct “no results” gating, and guarding single-token fuzzy matches from hijacking the queue.

<sup>Written for commit 4b80dcd0a712e8abcd91ad5082b2a70247611c58. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/2052?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Improved artist search with fallback support across Spotify, YouTube, and SoundCloud.
  - Playback now automatically selects the appropriate search method for results from alternate platforms.
  - Artist matching better prioritizes exact matches while retaining suitable partial matches.

- **Bug Fixes**
  - Searches continue after an individual provider fails or returns no results.
  - Improved handling when no qualifying artist matches are found.


LucasSantana-Dev:
## Second fix added: the artist filter was being discarded

Reported today — `/artist queen` played Bohemian Rhapsody and then **"Queencard" by i-dle**. Different failure from the one this PR opened for.

```ts
byArtist.length >= 3 ? byArtist : searchResult.tracks
```

The command computed exact / word / substring author matches and then threw all of them away unless one tier had **at least three** hits. Spotify's results for "queen" include tracks by i-dle, ABBA and others, so when fewer than three results were actually by the named artist, the filter was dropped and the provider's raw ranking was queued.

### Confirmed from production, not inferred

| Evidence | Value |
|---|---|
| `track_history` 19:46:14, Bohemian Rhapsody | `isAutoplay = false` |
| autoplay markers in that guild's logs | **none** |
| Queencard start | 19:46:21, **7s** after |

Both tracks came from one command, and Queencard can only reach that queue via the raw-results branch — so `byArtist.length < 3`.

### Change

Thresholds become `> 0`: take the narrowest tier that matched anything. Two tracks genuinely by the artist beat ten by whoever the provider ranked highly, because `/artist` promises tracks **by** the artist. The raw fallback is kept for the zero-match case so misspellings and odd author metadata still return something.

2 tests added, the first verified failing before the change. Full bot suite **3162 passed**.

Worth being explicit that these are two independent bugs in one command: the original fix covers **Spotify returning nothing**, this one covers **Spotify returning results for the wrong artists**. The first would not have prevented today's report.

LucasSantana-Dev:
P3 confirmed and fixed. Worth recording how, because my first check said the opposite.

I probed `addTrack.mock.calls` and got `["Someone Else","Third Party"]` — no Prince Royce, which looked like the comment was right and the finding wrong. Then I probed the other channel:

```
play:     https://example.com/Darte%20un%20Beso     <- Prince Royce
addTrack: ["Someone Else","Third Party"]
```

The first track goes to `play()`, not to the queue. So Prince Royce **is** surfaced first, `addTrack` never saw it, and the assertion counting `addTrack` calls could not have verified the claim the comment made.

Comment now says what the test actually guards — the fuzzy `>= 3` threshold — and records that the raw fallback still leads with the wrong artist here.

## The part worth a second look

That is arguably the residual gap this PR is about. With `exactMatch` empty and both fuzzy tiers under 3:

```ts
const byArtist = exactMatch.length > 0 ? exactMatch
    : wordMatch.length >= 3 ? wordMatch
    : substringMatch.length >= 3 ? substringMatch
    : []
const tracks = (byArtist.length > 0 ? byArtist : searchResult.tracks).slice(0, limit)
```

`byArtist` is empty, so the raw provider order wins — and the provider ranked Prince Royce first precisely because it fuzzy-matched "Prince". The threshold stops the filter from *trusting* one fuzzy hit, but the fallback then hands that same track the lead position anyway.

Same shape as `/artist queen` playing Queencard. Not changing it here since it is a behaviour decision, not a comment fix, but flagging it: dropping the single fuzzy match from the raw fallback (rather than falling back to the unfiltered list) would close it.

9 tests pass.

Issue #2054, opened 2026-08-21
fix(shared): log lines carry no level token, so Loki cannot label them


LucasSantana-Dev:
## Summary

`LogService.formatMessage()` emits `[timestamp] message` and nothing else. The severity is conveyed **only by ANSI color**, which no log shipper can read. Consequence: in the homelab Loki, `level="INFO"` and `level="WARN"` return **zero rows over 7 days** for `lucky-bot`, while `level="ERROR"` returns 1429 that are almost entirely wrong.

## Evidence

Code — `packages/shared/src/utils/general/log/service.ts:61-73`:

```ts
private formatMessage(params: LogParams): string {
    const { message, correlationId } = params
    let formattedMessage = message
    if (this.config.enableTimestamp) {
        formattedMessage = `[${new Date().toISOString()}] ${formattedMessage}`
    }
    if (this.config.enableCorrelationId && correlationId) {
        formattedMessage = `[${correlationId}] ${formattedMessage}`
    }
```

No level token. `getColor(level)` at `:120` is the only thing that varies by severity.

Loki, 7d, `container_name="lucky-bot"`:

| Query | Count |
|---|---|
| `{level="INFO"}` | **0** |
| `{level="WARN"}` | **0** |
| `{level="FATAL"}` | 0 |
| `{level="ERROR"}` | 1429 |

Of those 1429, **1417** are yt-dlp's own stderr, which happens to print a literal `ERROR:` prefix of its own. The remainder are false positives (see below).

## Why this matters now

Warn-level diagnostics were just added to `/artist` (#2052) and `/album` (#2053) precisely so a repeat failure is visible. **They are not queryable by level** — `{container_name="lucky-bot",level="WARN"}` returns nothing. Any Grafana alert built on `level` is simultaneously ~99% noise and blind to every warning. This compounds #2051 (warn never reaches Sentry either), so warn-level output currently has no reliable destination at all.

## Proposed fix (Lucky half)

Emit the level as a parseable token, e.g. `[2026-08-21T16:20:09.424Z] [WARN] Primary search failed...`, keeping the color for humans. Position it consistently so a shipper can anchor on it rather than substring-match.

Coordinate with the promtail half: LucasSantana-Dev/homelab#<promtail issue> — its regex substring-matches the level anywhere in the line, which is the other half of this bug and the source of the false positives.

## Repro

```bash
ssh homelab
[...]

PR #2059, opened 2026-08-21
fix(shared): write the log level as a parseable token

LucasSantana-Dev:
Refs #2054. Cross-repo counterpart: LucasSantana-Dev/homelab#418.

## Problem

Log lines carried no level token. Severity existed only as an ANSI colour (`getColor`, `service.ts:77-92`), which no shipper can read.

7d Loki, `container_name="lucky-bot"`:

| Query | Rows |
|---|---|
| `{level="INFO"}` | **0** |
| `{level="WARN"}` | **0** |
| `{level="ERROR"}` | 1429 |

Those 1429 were matched on *content*, not severity — promtail substring-searches for the keyword anywhere in the line. 1417 were yt-dlp's own `ERROR:` stderr prefix; the rest were a track titled **"10 BARRE DI TERRORE"**, whose name contains `ERROR`, labelling routine playback as an error.

Two Grafana panels consume this (`homelab .../lucky-app-metrics.json:383,407`, `level=~"ERROR|WARN|FATAL"`): the bot panel is ~99% noise and the backend panel has never shown a row.

## Change

Prepend `[LEVEL]` at index 0.

It is prepended **last**, after timestamp and correlationId, so its position does not depend on `enableTimestamp` or `enableCorrelationId`. That lets a shipper anchor on `^\[LEVEL\]` rather than scanning the line — and anchoring is precisely what stops message content from forging a level.

- `SUCCESS` maps to `INFO`: it is an informational outcome, and promtail/Grafana have no SUCCESS bucket.
- An out-of-range level falls back to `INFO` instead of emitting `[undefined]`, so the index-0 token is always parseable. `log(99)` is a real path — `service.spec.ts:70-79` exercises it.

Format goes from `[ts] msg` to `[WARN] [ts] msg`.

## Sequencing — this must deploy before the promtail change

Not interchangeable:

- **This first:** promtail's *current* substring regex immediately finds `[WARN]` in the line, so WARN starts labelling correctly. The "TERRORE" false positives persist until homelab#418 lands. Strictly better at every moment, no regression window.
- **homelab#418 first:** its anchored regex expects a token that does not exist yet, so `level` goes 100% empty and both Grafana panels go blank — losing the one signal that currently works.

## Tests

`levelToken.spec.ts`, 9 cases, **all 9 verified failing before this change**:

- a token per level, matching the anchored expression promtail will use
- `SUCCESS` maps to `INFO`
- out-of-range level falls back to `INFO`, never `[undefined]`
- token stays at index 0 with both prefix flags **off**, and with a correlationId present — the regression guard for the anchor
- the "BARRE DI TERRORE" case: content cannot forge a level

Full suites: shared **1420**, bot **3154**, backend **1352**. Typecheck clean across all four packages.

## Not covered here

Warn still never reaches Sentry (#2051) — `LogService.warn()` only calls `addBreadcrumb`. Separate fix.

---
## Summary by cubic
Writes a parseable [LEVEL] token at index 0 of every log line so shippers can anchor severity. Previously levels relied on ANSI color or substring matches, which mislabeled INFO/WARN and let message content forge ERROR.

- Adds `[LEVEL] ` outside color and appended last; stays at index 0 regardless of timestamp/correlationId; prefixes every physical line of messages, data, and error stacks; drops trailing empty segments.
- Sanitizes once per line in the prefixer, preserving JSON indentation and preventing log injection.
- Makes serialization robust: values that stringify to undefined fall back to String(), and if String() throws we emit "[unserializable]"; reads Error name/message/stack inside a try block to survive hostile getters.
- Maps SUCCESS to INFO; unknown levels fall back to INFO; never emits "[undefined]". No changes to `LogService` callers.

- Rollout
  - Deploy this before the anchored regex change in `LucasSantana-Dev/homelab#418`; anchoring first would empty the level label.
  - No migration required for log callers.

<sup>Written for commit 6f77b3c12d0b008a794495f74ecce7820ac080c1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/2059?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Log messages now include consistent, shipper-readable severity prefixes.
  - Success messages use informational severity.
  - Prefixes apply consistently across multiline messages, data, errors, and timestamped logs.

- **Bug Fixes**
  - Invalid levels safely fall back to informational severity.
  - Colored output keeps prefixes readable and correctly positioned.
  - Unserializable or hostile values no longer interrupt logging.

- **Tests**
  - Added coverage for multiline output, error stacks, fallbacks, hostile values, and level-like message content.


LucasSantana-Dev:
Resolving the two CodeQL threads: the alerts behind them are already closed.

```
gh api "repos/LucasSantana-Dev/Lucky/code-scanning/alerts?ref=refs/heads/fix/log-level-token&state=open"
[]
```

Both comments date from 01:44, before the commit that moved sanitisation inside `prefixLines`. The later scan cleared them; only the PR threads lagged. Nothing was suppressed and no suppression comment was added.

For the record, the log-injection property now holds at a single point:

```ts
return lines
    .map((line) => token + color(sanitizeForLogging(line)))
    .join(n)
```

Every logged line passes through `sanitizeForLogging` after the split, so no caller can forget it and no serializer can bypass it. The earlier shape sanitised at each caller and merely concatenated here, which is what broke the taint path CodeQL follows.

## The three real defects this PR went through

All mine, all the same class — guard the value, forget that touching the value can throw:

1. `JSON.stringify` returns `undefined` for functions and symbols, so `prefixLines` called `.split` on `undefined`.
2. The `?? String(x)` fallback itself throws on a null-prototype object with no `toPrimitive`.
3. The `Error` branch sat outside the `try`, so a throwing `stack` getter crashed the call. `instanceof Error` passes for such an object, so the check above gave no protection.

Each fix is verified by reverting it and watching its own test fail.

```
38 log tests, 1430 shared tests, tsc clean
```

PR #2071, opened 2026-08-21
fix(bot): drop the autoplay arm that calls a removed spotify endpoint

LucasSantana-Dev:
Found while investigating the `Lucky API Dependencies Down` alert. That alert was half wrong, and this is the half that is real.

## Measured, with Lucky own Spotify credentials

Ran from inside `lucky-bot`, so the results describe this app, not another:

```
v1/search                    HTTP 200
v1/recommendations           HTTP 404
v1/audio-features            HTTP 403
v1/artists/{id}/top-tracks   HTTP 403
```

Last.fm is **not** down, contrary to the alert. All five read methods Lucky uses return 200: `track.getInfo`, `artist.getSimilar`, `track.getSimilar`, `tag.getTopTracks`, `artist.gettoptags`.

## What that meant in practice

`collectSpotifyRecommendationCandidates` ran on every autoplay replenish: fetched a token, resolved up to 3 seed tracks, built `min_energy`/`max_danceability` constraint params, called a dead endpoint, swallowed the failure and returned `[]`. Pure no-op work on every cycle, invisible because `logAndSwallow` never surfaced it.

`packages/shared/src/utils/spotify/artistApi.ts:117` already migrated off this API in 2024 and carries a comment explaining the deprecation. `packages/bot` never did.

## What survives

The other two arms are unaffected and both measured healthy:

- seed search via `searchSeedCandidates` -> `v1/search` (200)
- Last.fm via `collectLastFmCandidates` (200 on every method)

So autoplay keeps working; it just stops doing one round trip that could not succeed.

## Tests kept, not deleted

The specs that mocked the removed arm used it as a convenient **candidate source**, not as the thing under test. They now drive the same assertions through `searchSeedCandidates`, so downstream coverage (scoring, dedup, duration limits) is unchanged.

The `integrates Spotify recommendations into queue` case is retargeted at the surviving arm and now asserts on the queue rather than on a specific API call, so it stays valid as arms are added or retired.

One case chained `mockResolvedValueOnce` on a shared token mock with a position reserved for this arm, so removing the arm shifted the chain by one and broke an unrelated audio-features assertion. That entry is gone. Worth noting the fragility: a test that depends on the call order of a shared mock breaks whenever a caller is added or removed.

## Verification

```
jest    3145 passed, 1 skipped, 241 suites
tsc     clean
eslint  0 errors (18 pre-existing warnings)
```

## Deliberately out of scope

`/v1/audio-features` is 403 for this app too, so `currentFeatures` is always null. `candidateScorer` degrades safely (`if (!currentFeatures || !userId) return tracks`), so nothing crashes, but 13 references across 7 files are now dead weight. That is its own change, filed separately.

`top-tracks` at 403 is not on Spotify deprecation list, which suggests the app is in Development Mode. Unconfirmed: the container `wget` does not return the error body.

---
## Summary by cubic
Drops the autoplay path that called Spotify’s removed `/v1/recommendations`, eliminating silent no-op work on every replenish. Autoplay behavior is unchanged: candidates still come from Spotify seed search and Last.fm, with fewer network calls.

- Delete `getSpotifyRecommendations` from `packages/bot/src/spotify/spotifyApi.ts` and its tests/retry specs; delete `collectSpotifyRecommendationCandidates` and its spec in `packages/bot/src/services/musicRecommendation/autoplay/spotifyRecommender.ts`.
- Candidate collection uses `searchSeedCandidates` only; tests now assert on queue outcomes instead of specific API calls, remove a duplicate test introduced by a rename, and drop one brittle token-mock chain entry.
- Verified: `v1/search` 200; Last.fm reads 200; 3145 tests pass; `tsc` clean; `eslint` clean.
- No migration for the bot. If any external code used `getSpotifyRecommendations`, switch to the seed-search flow.
- Out of scope: `/v1/audio-features` 403 keeps `currentFeatures` null; follow-up planned.

<sup>Written for commit 25a30337e9d289125e7a9c0c3b8ccf21fc1d2568. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/2071?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LucasSantana-Dev:
Both cubic findings were correct and are fixed. The graphify one I am pushing back on.

## P2 — the assertion did not verify the arm it named (fixed)

Right, and worse than a naming problem. `searchSeedCandidates` searches through `queue.player.search` with `SPOTIFY_SEARCH`; it never touches `spotifyApi.searchSpotifyTrack`. That mock was reachable **only** through the audio-features path.

So the assertion proved nothing about seed search, and it broke the moment that path was removed on the follow-up branch. I had introduced it in this PR while the comment directly above it claimed the opposite: "asserts on the queue, not on a specific API call". It now actually does that.

## P3 — the rename created a duplicate (fixed)

Also right. Renaming `should collect candidates from Spotify API` to `...from the seed search arm` made it identical to the pre-existing `should collect seed-based search candidates`: same setup, same mock, same call, only assertion strictness differing. Dropped the renamed one; the original stays.

## graphify: `searchSeedCandidates()` fans out to 7 callees

Not acting on this one. The function is **unchanged by this PR** — not a line of it moved. What changed is that its sibling `collectSpotifyRecommendationCandidates` was deleted, so `searchSeedCandidates` went from one of two exported functions in the module to the only one. The coupling delta is measuring the file, not the function.

Splitting a function nobody touched, to offset a metric that moved because a dead sibling was removed, would add indirection for no reader benefit. If that fan-out is worth addressing it should be its own change, judged on its own merits.

3153 tests pass.

Issue #1279, opened 2026-06-09
bug(bot): add timeout/AbortSignal to 9 Spotify fetch calls in spotifyApi.ts


LucasSantana-Dev:
**Phase**: 1 — Critical & High | **Task ID**: T3

## Evidence
- `packages/bot/src/spotify/spotifyApi.ts` L184, 248, 298, 377, 433, 477, 522, 530, 593 — direct `fetch()` without signal/timeout
- `withSpotifyRetry` handles 429 Retry-After but provides no overall deadline; a hung Spotify response blocks the caller indefinitely
- Bot-side sibling of closed backend issue #1203 (same bug class, different package)

## Acceptance
- [ ] Every fetch() in spotifyApi.ts carries an AbortSignal with a bounded timeout (~5-10s)
- [ ] withSpotifyRetry still handles 429 Retry-After correctly with the timeout in place
- [ ] Existing spotifyApi specs pass; a test covers the timeout/abort path

## Suggested approach
Add a fetchWithTimeout helper (AbortController, default 5s) or reuse the shared withTimeout util; thread the signal through all 9 call sites.

---
Created from `.claude/backlog/2026-06-09.md` by `/plan-to-issues`.
<sub>Surfaced by `/backlog` on 2026-06-09. Dedup key: `audit-deep:spotify-fetch-no-timeout:packages/bot/src/spotify/spotifyApi.ts`.</sub>

PR #808, opened 2026-05-05
feat(bot): add Spotify API 429 retry with Retry-After header

LucasSantana-Dev:
## Summary
- Adds retry logic to Spotify API calls when rate-limited (HTTP 429)
- Respects the \`Retry-After\` header value when provided, defaults to 1 second
- Retries up to 2 times before surfacing the error with a warning log

## Test plan
- [ ] Run \`npm run test --workspace=packages/bot\` — all tests pass
- [ ] Verify Spotify API calls retry on 429 responses in integration

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<h3>Greptile Summary</h3>

- Introduces `withSpotifyRetry` (up to 2 retries, honours `Retry-After`) wrapping 5 of 7 Spotify `fetch` call sites; `searchSpotifyTrack` and `getUserTopArtistsAndTracks` are still called directly without retry coverage.
- The retry catch block is dead code in production: the Web Fetch API always *resolves* for HTTP errors (including 429) rather than throwing, so `withSpotifyRetry`'s `catch` is never reached; the tests pass only because they mock `fetch` to throw a `Response`, which is not how the real API behaves.
- `Retry-After` header parsing uses `parseInt` on the raw header value; if Spotify sends an HTTP-date string (valid per RFC 7231) rather than a delta-seconds integer, `parseInt` returns `NaN`, making `delayMs = NaN` and `sleep(NaN)` collapses to a 0 ms delay.

<h3>Confidence Score: 3/5</h3>

Not safe to merge — the core retry feature is dead code at runtime and the tests validate an unrealistic mock path.

A P1 defect (retry catch is unreachable because fetch resolves on 429) means the entire feature ships as non-functional code. The NaN delay on HTTP-date Retry-After is an additional P1 edge case. Multiple P1s on the central changed path push the score to 3/5.

packages/bot/src/spotify/spotifyApi.ts — every withSpotifyRetry callback must explicitly throw the response on 429 before the generic !res.ok guard; packages/bot/src/spotify/spotifyApiRetry.spec.ts — tests must be rewritten to mock fetch resolving with a 429 Response rather than throwing one.

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
[...]

PR #2074, opened 2026-08-21
fix(bot): drop the audio-features scoring path, spotify returns 403

### LINKED ISSUE #2072: Spotify audio-features returns 403: currentFeatures is always null across 7 files
Measured from inside `lucky-bot`, with Lucky own Spotify credentials:

```
v1/search                    HTTP 200
v1/recommendations           HTTP 404   -> removed in #2071
v1/audio-features            HTTP 403   <- this issue
v1/artists/{id}/top-tracks   HTTP 403
```

## Effect

`getBatchAudioFeatures` and `getAudioFeatures` can never succeed, so `currentFeatures` is always `null`.

Nothing crashes. `candidateScorer.ts:644` degrades safely:

```ts
if (!currentFeatures || !userId) return tracks
```

But that means the energy/valence/danceability/tempo/acousticness scoring path is dead code that still costs a token fetch and a request on every replenish, and the autoplay ranking silently loses a signal it was designed around.

## Reach

```
13 refs  autoplay/candidateScorer.ts
 8 refs  candidateFallback.ts
 6 refs  spotify/spotifyApi.ts
 3 refs  autoplay/candidateCollector.ts
 3 refs  autoplay/audioFeatures.ts
 2 refs  autoplay/replenisher.ts
```

Plus `audioFeatures.ts` maintains a 5000-entry LRU cache with a 24h TTL that can only ever cache `null`.

## Options

1. **Remove the path**, as #2071 did for recommendations. Honest, and the scorer already handles its absence. Loses the capability if access is ever restored.
2. **Apply for extended access** on the Spotify app. `top-tracks` also returning 403 suggests Development Mode rather than the 2024 deprecation, since that endpoint is not on the deprecation list. Unconfirmed: the container `wget` does not return the error body, so the reason string was not read.
3. **Replace the signal.** Last.fm tags are already fetched (`getArtistTags`) and are healthy. Tag-based similarity would not reproduce tempo or acousticness, but it would restore a genre signal the scorer could use.

Option 2 should be checked first, because it is the only one that is free if it works. Read the 403 body with a client that returns it, and check the app access level in the Spotify dashboard.

Blocked on that answer before choosing between 1 and 3.

Related: #2071 (same investigation, the recommendations half).

LucasSantana-Dev:
Closes #2072. Replaces #2073, which GitHub auto-closed when #2071 merged with `--delete-branch` and took its base branch with it. Same content, rebuilt on current `main`.

## Measured, with Lucky own credentials from inside lucky-bot

```
v1/search / artists/{id} / albums/{id} / tracks/{id}   HTTP 200
v1/audio-features                                      HTTP 403
```

So `currentFeatures` was always `null` and `enrichWithAudioFeatures` returned its input unchanged on the first line. The 5000-entry LRU cache behind it, with a 24h TTL, could only ever hold `null`.

## The Last.fm replacement already exists

The plan was to substitute the Spotify signal with Last.fm tags. That work was already done: `candidateScorer` scores genre-family overlap in-pass via `calculateGenreFamilyPenalty` on Last.fm tags, and the code says so at the old call site:

> Genre-family scoring moved to in-pass `calculateRecommendationScore` via Last.fm tags so it works without a Spotify token. This pass is now limited to audio-feature (energy/valence) deltas, which the Last.fm-tag path cant substitute for.

What remained is energy, valence, tempo and acousticness. Last.fm does not expose those, so there is nothing to substitute with.

## Extended access does not bring them back

Ruled out with sources on #2072. Since 2025-05-15 it requires a registered business entity and 250k MAUs, and the deprecation notice says extended mode **preserved** these endpoints for apps that already had it on 2024-11-27. It is not a door that opens later.

## Removed

`enrichWithAudioFeatures` existed **twice**, in `candidateScorer.ts` and in `candidateFallback.ts`. Both are gone, along with `audioFeatures.ts`, `getAudioFeatures`, `getBatchAudioFeatures`, the `SpotifyAudioFeatures` / `SpotifyAudioFeatureConstraints` types, and the `currentFeatures` parameter threaded through the collector chain.

```
15 files changed, 29 insertions(+), 1164 deletions(-)
```

## Tests retargeted, not deleted

Two regression tests used the removed fetchers only as a **vehicle** to exercise `spotifyFetch` shared `AbortSignal` (#1279) and its 429 retry (PR #808 / Greptile). Those behaviours still matter, so both now run against `searchSpotifyTrack` and carry a comment saying why.

## Verification

```
jest    3122 passed, 1 skipped, 243 suites
tsc     clean
```

---
## Summary by cubic
Removes Spotify audio-features–based scoring because the endpoint now returns 403. Previously we boosted/penalized candidates using energy/valence/tempo/acousticness; in practice this path was a no-op (always null), so behavior does not change and scoring continues to rely on Last.fm tag–based genre-family checks.

- Delete the enrichment path: remove `enrichWithAudioFeatures` (from candidate scorer and fallback) and `audioFeatures.ts`; drop `getAudioFeatures`, `getBatchAudioFeatures`, and the `SpotifyAudioFeatures`/`SpotifyAudioFeatureConstraints` types.
- Simplify call sites: remove the `currentFeatures` parameter from candidate collection and replenisher; delete the related cache and the export in queue manipulation.
- Keep infra tests: retarget AbortSignal deadline and 429-retry regressions to `searchSpotifyTrack`; update Retry-After tests to return a valid search payload and assert non-null results. Remove an unused `getArtistGenres` import flagged by CodeQL.
- No rollout or user migration needed; compile-time errors will surface any stale imports.

<sup>Written for commit 7bb2effd858f1c2784758279f57b292a64e62c0c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/2074?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LucasSantana-Dev:
Two of three fixed. Pushing back on the third.

## cubic P3 — the retry tests could not fail (fixed)

Correct, and the sharpest finding of the three. When I retargeted those two Retry-After tests at `searchSpotifyTrack`, I changed the function but left the old audio-features payload:

```js
JSON.stringify({ id: "t", energy: 0.5, valence: 0.5 })
```

`searchSpotifyTrack` reads `data.tracks.items[0].id`, so that parses to `null`. The tests passed only because they assert `attemptCount === 2` and never look at the result. The retry could have returned anything and they would have stayed green.

Payload is now a search response, and both tests additionally assert the promise resolves non-null. Verified the assertion actually bites by reverting the payload:

```
2 failed, 5 passed
```

Before the change, that same revert was green.

## CodeQL — unused import (fixed)

`getArtistGenres` was left imported in `candidateFallback.ts` after `enrichWithAudioFeatures` came out. Removed. `eslint --quiet` and `tsc` are both clean.

## graphify — `getArtistPopularity()` coupling Ca·Ce = 12

Not acting on it, same reasoning as #2071. The function is untouched by this PR: not a line of it changed. What changed is that `spotifyApi.ts` lost 123 lines when the two audio-features fetchers were deleted, so the ratio moved because the denominator shrank. The metric is describing the file, not the function.

Refactoring a function nobody edited, to offset a number that moved because dead code was removed, adds indirection with no reader benefit. If that coupling is worth addressing it deserves its own change, judged on its own merits.

3122 tests pass.

PR #2081, opened 2026-08-22
fix(shared): write the log level as a parseable token

LucasSantana-Dev:
Reaberto: mergeStateStatus travado em BLOCKED com todas as condições satisfeitas.

---
## Summary by cubic
Writes a parseable [LEVEL] token at index 0 of every emitted log line so shippers can anchor severity. Previously severity was color-only, so Loki labels were empty and message/stack content could forge “ERROR”.

- Adds `LEVEL_TOKEN` (maps SUCCESS→INFO; out-of-range→INFO) and `prefixLines` to prefix and sanitize per physical line; sanitizes the single-line message before splitting to block forged records.
- Error formatting: reads `name`, `message`, and `stack` once; strips the message from the stack only when a header exists, including when the first stack line equals the message; requires indentation for frames (`/^\s+at\s/`) to avoid frame-shaped headers/names; keeps headerless stacks intact; removes the trailing empty token line when no stack.
- Hardens serialization and fallbacks: guards `JSON.stringify` returning undefined or throwing, hostile Error getters, and circular/throwing `toJSON`; `toError` reuses `serializeData`; provides a safe “[unserializable]” fallback.
- Keeps the token outside the color wrapper; holds index-0 position with timestamp and correlationId on or off; prefixes every line of pretty-printed JSON data.

**Rollout**
- Deploy `packages/shared` before updating promtail to an anchored `^\[LEVEL\]` regex; anchoring first will empty the level label.

<sup>Written for commit 329d0a32920651682efd01aa4684afd0f18fd268. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/2081?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved log output with consistent, parseable severity tokens on every line.
  - Preserved severity tokens across colored, timestamped, correlated, and multiline logs.
  - Improved handling of invalid, unserializable, undefined, or unexpected values without interrupting logging.
  - Prevented message content, control characters, and forged stack frames from spoofing or corrupting log records.
  - Safely handled errors with modified messages, names, or throwing properties.

- **Tests**
  - Added coverage for error formatting, serialization resilience, multiline messages, forged stack frames, and severity-token behavior.


LucasSantana-Dev:
Resolving the three remaining threads. Two are stale, one is a false positive with the measurement below.

## The two CodeQL threads (15:58, 17:10)

No live alert stands behind either:

```
gh api "repos/LucasSantana-Dev/Lucky/code-scanning/alerts?ref=refs/heads/fix/log-level-token&state=open"
[]
```

Both predate the commit that moved sanitisation inside `prefixLines`. The later scan cleared them; only the threads lagged. Nothing suppressed, no suppression comment added.

## The `data` sink thread (16:10) — measured false

The claim is that a line break in serialized `data` is preserved as another log record. `JSON.stringify` escapes a newline inside a string as the two characters backslash-n, so it never becomes a real line break:

```
JSON.stringify({user: "foo\nbar[ERROR] forged"}, null, 2)

{
  "user": "foo\nbar[ERROR] forged"
}

physical lines: 3        <- indentation only, ours
a line starting with "bar"? false
```

The only real newlines in that output are the indentation we asked for, and every one of those lines gets the level token. No user-controlled newline reaches the sink as a record boundary. No change made.

The message path was a different story and that one was real — fixed in the commit that sanitises `formattedMessage` before the split.

## Where this PR ended up

Nine review rounds on `serializeError`, every one a real defect except the `data` claim. The last round is worth noting: the reviewer suggested recognising one more header form, which would have been the fourth consecutive special case. Looking for what actually separates a header from a frame gave a one-character answer — V8 indents every frame and never indents the header, so `/^\s*at\s/` became `/^\s+at\s/` and three branches disappeared.

Six fixtures cover it now: normal, headerless with a message of `at `, empty name with a frame-shaped message, empty message, a frame ending with the message, and a frame-shaped name. None forges a record, none is corrupted.

```
50 log tests, 1442 shared tests, tsc clean
```

PR #2248, opened 2026-09-06
fix(bot): log missing-permissions failures on automod message delete

### LINKED ISSUE #1919: fix(bot): message.delete() swallows permission errors as if they were 404s
Good first issue: two files, one line each, and the existing logging helpers are already imported in both.

## The problem

Both auto-moderation paths delete offending messages like this:

`packages/bot/src/handlers/message/spamHandler.ts:41`
```ts
await message.delete().catch(() => {})
```

`packages/bot/src/handlers/message/autoModHandler.ts:140`
```ts
await message.delete().catch(() => {})
```

That empty catch treats every failure as equally uninteresting. Two very different things end up silent:

- **`10008 Unknown Message` / HTTP 404** — someone else already deleted it, or it aged out. Genuinely expected, and correctly ignored.
- **`50013 Missing Permissions`** — the bot lost `Manage Messages` in that channel. Auto-mod is now **silently doing nothing**: it detects spam, tries to delete, fails, and reports success by saying nothing at all.

The second case is a real outage of a feature the server owner thinks is on, and there is currently no signal anywhere. Not in the logs, not to the operator.

## The task

Distinguish the two. Keep ignoring the "already gone" case; log anything else.

Both files already import a logger (`errorLog` in the surrounding `catch`), so you have a pattern to follow a few lines away. discord.js throws `DiscordAPIError` with a numeric `.code`, so something along these lines:

```ts
await message.delete().catch((error) => {
    // 10008 = Unknown Message: already deleted or aged out. Expected.
    if (error?.code === 10008) return
    warnLog({ message: 'Failed to delete message', error, data: { channelId: message.channelId } })
})
```

Use whichever log level fits the repo's conventions — check how neighbouring handlers classify recoverable-but-notable failures.

## Worth deciding while you are in there

Both call sites want identical behaviour, so a small shared helper is probably better than the same catch block twice. Your call, and worth saying which you chose in the PR.

## Verifying

- `npm run test:bot` stays green.
- A test that a 404-shaped error stays silent while a permissions-shaped error logs would be a genuinely useful addition, and both files already have specs to extend.

## Background

Split out of #1792, which raised this pattern but pointed at `ttlDeleteHandler.ts` — a file that was never merged, because TTL delete mode got deferred to #1813. The concern was right; it just lives in these two files instead.

LucasSantana-Dev:
## Summary
- spamHandler and autoModHandler both swallowed every message.delete() failure identically via an empty catch
- 50013 Missing Permissions is a real outage (automod detects a violation, tries to delete, silently fails) with zero signal anywhere
- Added a shared safeDeleteMessage() helper: ignores 10008 Unknown Message (expected, already gone), warnLog()'s anything else with channelId

## Test plan
- [x] npm run test:bot (spamHandler.test.ts, autoModHandler.test.ts) — new regression tests for both the 10008-silent and 50013-logged cases
- [x] tsc --noEmit clean

Closes #1919

---
## Summary by cubic
Fixes automod and spam handlers silently swallowing message delete failures, so missing-permission errors (50013) are now logged with the channel ID instead of vanishing. Already-deleted messages (10008) stay silent. Closes #1919.

**Bug Fixes**
- Logging is throttled per guild and can't affect the delete outcome; if the logger throws, the original error and logging failure fall back to stderr via `console.error`.
- Adds a shared `safeDeleteMessage()` helper and regression tests for both the silent and logged failure cases.

<sup>Written for commit 04868c6cc0de4ae70a2e6071dfba942bfc75d90c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/2248?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LucasSantana-Dev:
non-destructive-confirmed: this PR does not add a new destructive interaction. Both `await message.delete()` call sites already existed, unchanged, in spamHandler.ts and autoModHandler.ts before this diff - this PR only extracts the shared call into messageDelete.ts and changes what happens in the catch block (silent on 10008, warnLog on anything else). The gate's diff-scan flags the relocated call as an "added" line because it is textually new to that file, but no new delete capability or behavior change to the delete itself is introduced. No live smoke needed per the false-positive carve-out in the destructive-interaction merge-gate ADR.

PR #2249, opened 2026-09-06
chore(config): remove 5 dead vars from .env.example

### LINKED ISSUE #2093: chore(config): BOT_SUPPORT_SERVER and BOT_WEBSITE are declared in .env.example and read by nothing
Both variables are documented in `.env.example` and referenced nowhere in the codebase.

```
$ grep -rn "BOT_SUPPORT_SERVER\|BOT_WEBSITE" --exclude-dir=node_modules --exclude-dir=.git \
      --exclude-dir=dist --exclude-dir=coverage --exclude-dir=.claude .
.env.example:30:BOT_WEBSITE=
.env.example:31:BOT_SUPPORT_SERVER=
```

Two hits, both the declarations themselves. No consumer in `packages/bot`, `packages/backend`, `packages/frontend` or `packages/shared`.

## Why it matters

Dead config is worse than missing config. Someone self-hosting a fork reads `.env.example`, fills `BOT_SUPPORT_SERVER` in with their own server, and nothing happens. The file reads as a contract that the code does not honour.

It also cost real time here: while fixing #2087 I found `BOT_SUPPORT_SERVER` and briefly treated it as the existing wiring for the support-server URL, which would have been the reuse-what-is-here answer. It is not wired to anything, so #2087 got a shared constant instead.

## Two ways out, pick one

1. **Delete both lines** from `.env.example`. Smallest change, honest immediately.
2. **Wire them up.** `BOT_SUPPORT_SERVER` would naturally back `SUPPORT_SERVER_INVITE_URL` (`packages/shared/src/constants/support.ts`, added in #2092) and `BOT_WEBSITE` the site URL, so a fork could point both at its own infrastructure without patching source. This is the better answer for self-hosters, which the README leans on as a selling point, but it is more work and needs a decision about defaults when the variables are empty.

Worth checking the rest of `.env.example` for the same problem while in there.

## Reproduces on clean main?

Yes, pre-existing.

## Acceptance

- No variable in `.env.example` is documented without a consumer, or the ones that remain are explicitly marked as reserved.

LucasSantana-Dev:
## Summary
- BOT_SUPPORT_SERVER and BOT_WEBSITE (named in the issue) plus BOT_DESCRIPTION, BOT_AVATAR_URL, BOT_COLOR (found while checking the rest of the file, as the issue suggested) are documented in .env.example but have zero consumers in packages/bot, backend, frontend, or shared
- Deleted all 5 lines; kept BOT_NAME (consumed by docker-compose.staging.yml) and every other var, all of which have real consumers

## Test plan
- [x] grep confirms zero references to any of the 5 removed vars outside .env/.env.example
- [x] tsc --noEmit clean (no env schema referenced them)

Closes #2093

---
## Summary by cubic
Removes 5 unused env vars from `.env.example` (BOT_DESCRIPTION, BOT_AVATAR_URL, BOT_COLOR, BOT_WEBSITE, BOT_SUPPORT_SERVER) so self-hosters no longer see config options that do nothing, and drops the stray `messageDelete.ts` file carried in from #2248 so Sonar's coverage gate clears. Only `.env.example` changes remain in the diff; `BOT_NAME` and all other vars are kept because they have real consumers.

<sup>Written for commit 8cad8b818d494fd0cb9126bc12a5225fa8e785d7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/2249?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LucasSantana-Dev:
Destructive interaction gate false positive: message.delete() already existed in both autoModHandler.ts and spamHandler.ts on main before this PR. This PR only extracts the existing delete call into a shared safeDeleteMessage helper that logs failures other than 10008 Unknown Message; it does not introduce any new destructive capability or change what gets deleted, when, or under what conditions. Adding non-destructive-confirmed label per the destructive interaction merge gate ADR's override path.

PR #2310, opened 2026-09-08
fix(bot): delete session snapshot on voice-kick disconnect

### LINKED ISSUE #2309: fix(bot): voice-kick disconnect never deletes session snapshot, orphan monitor rejoins with stale queue
## Bug

Reported by operator: "The bot is connecting back every time I cut the connection. I clearly don't want the bot to come back and it returns with bad songs."

## Root cause

`setupVoiceKickDetection` (`packages/bot/src/handlers/player/lifecycleHandlers.ts:22-34`) only calls `musicWatchdogService.markIntentionalStop(guildId)` when the bot is disconnected from voice directly (not via `/stop` or `/leave`). It never calls `musicSessionSnapshotService.deleteSnapshot(guildId)`.

- The intentional-stop flag is intentionally temporary: it auto-clears after `timeoutMs + 10_000` (35s default, `watchdog.ts:130-134`), by design, so it doesn't block a legitimate future session.
- Meanwhile a session snapshot is saved unconditionally on every track start (`trackHandlers.ts:242`), so a fresh snapshot almost always exists on disk while music is playing.
- `startOrphanSessionMonitor` (`watchdog.ts:277-391`) polls every 60s. Once the intentional-stop flag clears (~35s after the kick) it finds: no playing queue, flag cleared, a snapshot <30min old, and members still in the voice channel → it rejoins and restores the stale snapshot, resuming the old queue/autoplay tracks the user just disconnected to get rid of.

Contrast with `/stop` (`stop.ts:29-33`), which already does `markIntentionalStop` **and** `deleteSnapshot` together — the kick-detection path never got the same treatment.

Related but distinct from #2241 (saveSnapshot resurrecting a `/stop`-deleted session) and #2246 (stale flag masking a new session's snapshot save) — both already fixed. This is the third, previously-uncovered snapshot-lifecycle gap: the voice-kick path itself never clears the snapshot.

## Fix applied

`packages/bot/src/handlers/player/lifecycleHandlers.ts`: `setupVoiceKickDetection`'s listener is now `async` and calls `await musicSessionSnapshotService.deleteSnapshot(oldState.guild.id)` alongside `markIntentionalStop`, mirroring `/stop`.

`packages/bot/src/handlers/player/lifecycleHandlers.spec.ts`: added `deleteSnapshotMock` to the mocked `musicSessionSnapshotService`, made the kick-detection test async, and asserted `deleteSnapshot` is called with the guild id.

Verified: `npx jest src/handlers/player/lifecycleHandlers.spec.ts` (29/29 pass), `npx tsc --noEmit` clean.

LucasSantana-Dev:
## Bug

Bot reconnects to voice on its own after being disconnected and resumes an old/unwanted queue ("bad songs").

## Root cause

`setupVoiceKickDetection` marked the disconnect as intentional but never deleted the session snapshot. That flag auto-clears after the watchdog timeout (~35s) by design, so it doesn't block a future session — but a snapshot is saved on every track start, so it almost always survives the kick. Once the flag clears, `startOrphanSessionMonitor`'s 60s poll finds a fresh snapshot + members still in the channel and rejoins, replaying the queue the user just disconnected to escape.

`/stop` already pairs `markIntentionalStop` with `deleteSnapshot`; the voice-kick path never got the same fix.

## Fix

`setupVoiceKickDetection`'s listener now also calls `deleteSnapshot` for the guild, mirroring `/stop`. Test updated to cover it.

Closes #2309

## Test plan
- [x] `npx jest src/handlers/player/lifecycleHandlers.spec.ts` — 29/29 pass
- [x] `npx tsc --noEmit` — clean

---
## Summary by cubic
When the bot is kicked from voice, it previously marked the stop intentional but kept the session snapshot, allowing the orphan-session monitor to rejoin later and replay the old queue. The handler now deletes the guild snapshot during disconnect, preventing stale-session recovery while preserving intentional-stop behavior.

- Adds regression coverage for awaiting snapshot deletion.
- Fixes #2309.

<sup>Written for commit a14adab3e6dfb402671bc2bd1aeda86ba9ae5115. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/2310?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LucasSantana-Dev:
Confirmed real, but a narrower and distinct race from the one this PR fixes: an in-flight `checkAndRecover`/`recoverOrphanSession` call that already passed its entry-time `isIntentionalStop` check could still complete after this `deleteSnapshot`. Fixing that requires re-checking the flag immediately before each connect/restore point in both functions — broader than this narrow bug-fix PR's scope. Tracked separately: #2311.

Issue #1275, opened 2026-06-09
fix(autoplay): rejection telemetry never records — acceptance rate is blind (0 rejected / 14d)


LucasSantana-Dev:
## Evidence (prod, last 14 days via scripts/autoplay-telemetry.sql)
187 recommendation picks → **68 accepted, 0 rejected, 119 pending**. Acceptance rate reads 1.000 across every source — because the rejection half is never captured. The metric is effectively blind to skips.

## Root cause
`packages/bot/src/handlers/player/trackHandlers.ts:361` records `outcome: 'rejected'` **only** when an autoplay track is skipped within `OUTCOME_REJECT_EARLY_SKIP_MS` (~5s). A skip after 5s but before the accept threshold is explicitly dropped as "ambiguous" (line ~370 comment: "do not record"). Real rejections — listener hears 10–20s of a wrong-genre/wrong track then skips — land in that dead zone and are never written. Over 14 days of active skip-testing this produced **zero** rejections.

Possible secondary factor: confirm the player `skip` event + `guildTrackStartTimes` are reliably set for autoplay tracks (if the handler/timing isn't firing, no skip is ever evaluated).

## Why it matters
Without a rejection signal, per-source/per-mode acceptance (the input for all autoplay tuning + the ADR 2026-06-22 revisit) is meaningless. We've been tuning the recommender with the instrument half-off.

## Acceptance criteria
- [ ] A skip before ~30% completion (not just sub-5s) records `outcome: 'rejected'` for autoplay tracks; only genuinely-ambiguous mid-skips are dropped (and document the band).
- [ ] Verify the skip path actually fires for autoplay tracks (guildTrackStartTimes set; event wired).
- [ ] After deploy, re-running scripts/autoplay-telemetry.sql shows a non-zero rejected count and a meaningful per-source acceptance spread.
- [ ] Unit test the accept/reject/ambiguous classification by elapsed-time band.

## Blocked by
None. Prereq for data-driven autoplay tuning (relates to ADR 2026-06-07 revisit).

LucasSantana-Dev:
## Re-opened — #1276 fix did not move the needle (7-day verify, #1459)

Verified 2026-06-17 against prod Postgres: `rejected` is still **0** (223 picks / 85 accepted / 138 pending in 14d; **0 rejected all-time across 269 rows**). Full read in #1459. SUCCESS criterion #1 failed.

### Ruled out
[...]

PR #2322, opened 2026-09-10
fix(bot): prevent track key collision for repeated plays

### LINKED ISSUE #2298: fix(bot): prevent track key collision for repeated plays in autoplayOutcomeTracking
When the same track repeats while terminal handlers overlap, trackStartKey (guildId::trackId) cannot distinguish the two play lifecycles. Use a per-play identifier or lifecycle-specific key so one repetition cannot overwrite and delete another repetition's outcome timing.

### LINKED ISSUE #2301: test(bot): remove tautological cache tests from autoplayOutcomeTracking.spec.ts
Two tests in autoplayOutcomeTracking.spec.ts only verify lru-cache's own get/set/delete behavior, not the module's logic. They don't invoke playerStart/playerSkip/playerFinish handlers, so the 30% threshold and reset-on-completion logic is not tested. This behavior is already covered in trackEventHandlers.spec.ts. Delete these tautological tests to avoid false confidence.

LucasSantana-Dev:
## Summary

Fixes key collision when the same track repeats while handlers overlap (#2298). 
When Track A plays, then plays again before the first completes, they need distinct cache entries so outcome ratios aren't corrupted. Also removes tautological cache tests (#2301).

## Root Cause

`trackStartKey(guildId, trackId)` returns an identical string for repeated plays of the same track. When playerStart fires for the second instance, it overwrites the first's start time in the cache, causing the first play's playerFinish to calculate outcome ratio against the wrong (newer) timestamp.

## Solution

Track the start time per (guildId, trackId) pair in `currentTrackPlayStart` Map. Use this timestamp as part of the cache key: `guildId::trackId::startTime`. Each play instance gets a unique entry that persists until its terminal event (playerFinish/playerSkip) deletes both the cache entry and the tracking entry. Handlers supply the start time when calling `trackStartKey()`, or it's looked up if not provided (backward-compatible fallback).

## Test Coverage

Removed two tautological test describe blocks from autoplayOutcomeTracking.spec.ts that only tested cache behavior (get/set/delete) without invoking handlers. Actual logic is covered in trackEventHandlers.spec.ts.

Closes #2298
Closes #2301

---
## Summary by cubic
Fixes cache key collisions that corrupted autoplay outcome ratios when the same track played again before its previous play finished, including repeat modes that reuse the same track instance (#2334).

- Start times live in a `WeakMap` keyed on the track object; each entry holds a FIFO queue so overlapping plays of the same instance each keep their own start time.
- `playerFinish` and `playerSkip` now take their start time synchronously at handler entry, before any await, so overlapping finishes can't take the same entry.
- Removed the string-keyed LRU cache, which was pure indirection over the map and could collide when two plays started in the same millisecond.
- Removed tautological cache-only tests from `autoplayOutcomeTracking.spec.ts`; real coverage lives in `trackEventHandlers.spec.ts`, which now proves overlapping same-instance plays, concurrent finishes, and two distinct track objects sharing one track id keep separate start times.

<sup>Written for commit 9ce27f14f29b03f9bfe47c09869de3fdcb57eae9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/2322?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Closes #2301


LucasSantana-Dev:
All three cubic findings were correct and are fixed in `b5764519`, which simplifies rather than patches:

- **Sub-millisecond key collision (P2):** the string-keyed LRU is gone entirely. It only ever held what the WeakMap already held, keyed by a string that embedded the same timestamp — so it was indirection that could collide. Reads/writes now go straight to the WeakMap via accessors. Object identity cannot collide, needs no TTL or size cap, and frees itself with the track.
- **Tautological test (P2):** removed. It set and read the WeakMap without touching our code. The interleaving it claimed to cover is already tested through the handlers by the existing #1275 probe.
- **Reset helper no longer clearing start times (P3):** the map is now module-local and reassigned by `__resetTrackHandlerCachesForTests`, so the reset works. That mattered: an existing test depends on start times being dropped before a finish event arrives.

Net -51 lines. 29/29 tests pass, typecheck clean.

PR #2323, opened 2026-09-10
fix: re-check intentional-stop before watchdog recovery actions

### LINKED ISSUE #2311: bot: watchdog recovery doesn't re-check intentional-stop right before connecting/restoring
## Bug (narrow race, found by cubic reviewing #2310)

`MusicWatchdogService.checkAndRecover` (`watchdog.ts:176-275`) and `recoverOrphanSession` (`watchdog.ts:310-391`) both check `this.intentionalStops.has(guildId)` once, at entry, then `await` multiple times (connection rejoin/ready wait, snapshot fetch, `queue.connect`, `restoreSnapshot`) before actually resuming playback — without re-checking the flag immediately before those final connect/restore calls.

If a recovery call is already past its entry check when a kick/intentional-stop happens (e.g. `setupVoiceKickDetection`, `/stop`, `/leave`), the in-flight recovery can still complete and rejoin/replay a session the user just intentionally ended.

This is a separate, deeper issue from #2309 (fixed in #2310): #2309 closed the dominant ~35s-window bug (snapshot survived past the intentional-stop flag's auto-clear, so a *later* orphan scan wrongly treated the session as abandoned). This issue is about an *in-flight* recovery racing the intentional-stop signal itself — much narrower window (sub-second, requires a scan tick to land mid-kick), but the same class of bug.

## Suggested fix

Re-check `this.intentionalStops.has(guildId)` immediately before each connect/restore action in both `checkAndRecover` and `recoverOrphanSession` (not just at entry), or have the kick-detection path cancel/await any in-flight recovery lock (`recoveryInProgress`) before proceeding.

## Reference

cubic comment on PR #2310: https://github.com/LucasSantana-Dev/Lucky/pull/2310#discussion (line 34 of `packages/bot/src/handlers/player/lifecycleHandlers.ts`)

LucasSantana-Dev:
## Root cause

`MusicWatchdogService.checkAndRecover()` and `recoverOrphanSession()` checked the intentional-stop flag only at entry, then awaited multiple times (connection rejoin/ready wait, snapshot fetch, queue.connect, play/restoreSnapshot) without re-checking the flag.

If a stop command or voice kick occurred during this window, the in-flight recovery would still complete and resume playback, contradicting the user's intent to stop the session.

## Fix

Re-check `this.intentionalStops.has(guildId)` immediately before:
- In `checkAndRecover`: before `queue.node.play()` calls (lines 235-248)
- In `recoverOrphanSession`: before `queue.connect()` and `restoreSnapshot()` calls (lines 353-363)

## Test evidence

Added two new tests:
- `checkAndRecover aborts if intentional stop is set mid-flight before play (#2311)` — verifies recovery skips play() when flag is set during connection rejoin/wait
- `recoverOrphanSession aborts if intentional stop is set mid-flight before connect (#2311)` — verifies orphan recovery skips connect/restore when flag is set

All 28 watchdog tests pass. Typecheck clean.

Closes #2311

---
## Summary by cubic
Fixes the watchdog recovery race where playback could resume after a user stopped the session mid-recovery. `checkAndRecover()` and `recoverOrphanSession()` now re-check the intentional-stop flag immediately before `play()`, `connect()`, and `restoreSnapshot()` calls, so in-flight recoveries abort when a stop command or voice kick arrives.

- Aborted orphan recoveries now delete the queue they created so the next scan doesn't mistake it for a live session.
- Adds three tests covering the mid-flight stop race across both recovery paths, including the early bail-out before `connect()`.

<sup>Written for commit 57bd5dabe1a6139ff9ac70a65455a7be07448e3b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/2323?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LucasSantana-Dev:
Fixed in `2d11382c`:

- **Leaked queue node (P2):** correct catch. Both early returns created the queue via `player.nodes.create(guild)` and then returned without it, leaving an unconnected node registered for the guild — which the next orphan scan would read as a live session. Both returns now tear down a queue *this call* created (never one it found), and the mid-flight test asserts `queue.delete()` runs. Verified the assertion fails with the teardown removed and passes with it.
- **Stale line numbers (P3):** removed the hardcoded references from the comment rather than updating them, since they go stale again on the next edit.

- **Deeper `restoreSnapshot` race (P1):** valid, and deliberately not fixed here. The re-check narrows the window but cannot cover work spanning awaits — a stop arriving while `restoreSnapshot()` is mid-flight still resumes playback. Closing it properly changes `restoreSnapshot`'s contract for every caller, which is beyond this PR's scope. Filed as #2335 with the three options.

28/28 tests pass.

PR #2325, opened 2026-09-10
fix(bot): apply anchored-suffix check to clean-title

### LINKED ISSUE #2312: bug(bot): cleanTitle strips real song titles containing a version keyword (e.g. "Live Wire")
## Bug

Found while fixing #2265's \`hasVersionMarker\` (cubic caught the same root cause there — see PR #2265, commit c11dd955). \`cleanTitle\` (\`packages/bot/src/utils/music/searchQueryCleaner.ts:204-238\`, on \`main\` already, unrelated to that PR's diff) has the identical bug in its own hyphenated-suffix loop:

\`\`\`ts
for (const sep of [' – ', ' - ', ' — ']) {
    const idx = cleaned.indexOf(sep)
    if (idx > 0) {
        const suffix = cleaned.slice(idx + sep.length).trim()
        if (isVersionSuffix(suffix)) {
            cleaned = cleaned.slice(0, idx)
            break
        }
    }
}
\`\`\`

\`isVersionSuffix\` (line 181) falls back to a non-anchored keyword match (\`VERSION_KEYWORD_RE.test(suffix)\`) for any suffix ≤40 chars. Since \`noiseTerms.json\`'s \`versionVariants\` includes short common words like \`live\`, \`clean\`, \`cover\`, \`demo\`, \`remix\`, a real title like **"Motörhead - Live Wire"** gets its \`- Live Wire\` suffix matched (because "live" appears as a whole word inside "Live Wire") and stripped down to just "Motörhead" — silently corrupting search queries for any song whose title happens to contain one of these common words after a hyphen.

This is worse than the \`hasVersionMarker\` case fixed in #2265: \`cleanTitle\` actually **mutates** the query text (used for the real search), rather than just gating a boolean.

## Suggested fix

Apply the same pattern used to fix \`hasVersionMarker\` in #2265: check \`HYPHENATED_VERSION_SUFFIXES.some(re => re.test(suffix))\` directly instead of \`isVersionSuffix(suffix)\`, and use \`lastIndexOf\` instead of \`indexOf\` for the same "final segment" reasoning. Verify with a case like \`cleanTitle('Motörhead - Live Wire')\` → should stay \`'Motörhead - Live Wire'\`, not get truncated to \`'Motörhead'\`.

Also double check whether \`isVersionSuffix\`'s broad fallback branch (the whole reason for both bugs) is still needed by any other caller before deciding whether to fix call sites individually or tighten \`isVersionSuffix\` itself.

LucasSantana-Dev:
## Root cause
`cleanTitle()` used `indexOf()` and `isVersionSuffix()` to detect version suffixes in hyphenated segments. The `isVersionSuffix()` helper falls back to non-anchored keyword matching for short suffixes (≤40 chars), which matches version keywords anywhere in the text. This caused false positives: real song titles like "Motörhead - Live Wire" were truncated because "live" matched as a version keyword.

## Fix
Applied the same anchored-suffix pattern from the `hasVersionMarker()` fix (PR #2265):
- Changed `indexOf()` to `lastIndexOf()` to check the final segment
- Replaced `isVersionSuffix()` with direct `HYPHENATED_VERSION_SUFFIXES.some((re) => re.test(suffix))` to avoid the keyword fallback

This ensures version keywords only match when they appear as complete suffixes after hyphenation, not as substrings within real song titles.

## Test evidence
- All 94 existing tests pass
- Added regression test: "Motörhead - Live Wire" now correctly stays intact (not truncated to "Motörhead")
- Updated multi-suffix test to reflect correct behavior: "Beyoncé - Halo - Versão Forró" → "Beyoncé - Halo" (strips final version suffix)

Closes #2312

---
## Summary by cubic
Fixes `cleanTitle()` so version keywords only strip when they appear as the final hyphenated suffix, preventing real titles like "Motörhead - Live Wire" from being truncated.

- Switches from `indexOf()` to `lastIndexOf()` and checks `HYPHENATED_VERSION_SUFFIXES` directly instead of `isVersionSuffix()`, which is now removed as dead code.
- Adds a regression test for "Motörhead - Live Wire" and updates the multi-suffix case to strip only the final version suffix.

<sup>Written for commit e32d97a064846a7ccb21be04fdb445cd60a9cf6e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/LucasSantana-Dev/Lucky/pull/2325?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LucasSantana-Dev:
Correct — fixed in `0cbc7163`. Replacing the call in `cleanTitle` removed `isVersionSuffix`'s last caller, so the helper is deleted. `VERSION_KEYWORD_RE` stays: it is still used by the query-level check further down the file. 94/94 tests pass.

